PR #2, opened 2019-08-29
remove build-related directories (public, resources)

landam:
Build-related directories (`public`, `resources`) should be removed from git.


lucadelu:
I fully support this pull request, tested and it works as expected

PR #12, opened 2019-12-31
History

lucadelu:
Added several pages about history

There are problems with links going to in a new line within the <li>, see the pictures
problem_link


neteler:
@lucadelu: please merge

lucadelu:
> @lucadelu: please merge

@neteler there is problem with links because they go in a new line.

@nbozon do you have idea how to fix this?

neteler:
@lucadelu pls merge, we need to move on. If @nbozon doesn't answer we cannot wait forever.

lucadelu:
> 
> @nbozon I found how to solve it, the problem is in the style.css file at this line https://github.com/OSGeo/grass-website/blob/master/themes/grass/assets/css/style.css#L731
> 
> I fix the problem removing the display:block, @nbozon is this a problem?

implementation in #24

nbozon:
+1 @lucadelu , you can probabaly use this too:

      display: inline-block 

So links will keep display as block element, but won't jump a new line

lucadelu:
> +1 @lucadelu , you can probably use this too:
> 
> ```
>   display: inline-block 
> ```
> 
> So links will keep display as block elements (needed for style), but won't jump on a new line

done in e318887 , merged pull request.

Now I have new problem, it seems that html code is not rendered anymore. I remember that pages I created like releases.md were working now it is completely blank.

Should we convert html code to markdown?

neteler:
@nbozon do you have a hint for us? thx

veroandreo:
Found it here: https://jdhao.github.io/2019/12/29/hugo_html_not_shown/

Starting in hugo 0.60, HTML tags are no longer displayed by default, hence we need to add this:

```
[markup]
  defaultMarkdownHandler = "blackfriday"
```

to `config.toml`. I tested locally and works fine

lucadelu:
I'm going to merge, improvements needed

PR #16, opened 2020-01-07
improved tutorials page

lucadelu:


neteler:
Thanks - I'll merge, from that we can continue to edit.

PR #42, opened 2020-03-28
Fix HTML tags display for hugo >0.60

veroandreo:
As explained here: https://jdhao.github.io/2019/12/29/hugo_html_not_shown/, starting from hugo version 0.60, the HTML tags are no longer displayed by default because hugo has changed the Markdown renderer from blackfriday to goldmark.

Hence, one of the proposed solutions is adding blackfriday back in the `config.toml` file as in this PR

See #12

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -21,6 +21,9 @@ greyColorDark = "#A0A0A0"
   changefreq = "monthly"
   filename = "sitemap.xml"
   priority = 0.5
+
+[markup]
+  defaultMarkdownHandler = "blackfriday"
   
 [Languages]
 


lucadelu:
@nbozon could we merge this?

mlennert:
As I mentioned in [1], my general question is how much html we are actually still supposed to use with Hugo. Normal language for Hugo is Markdown IIUC, so shouldn't our first response be to replace all this HTML by markdown syntax ?



[1] https://lists.osgeo.org/pipermail/grass-dev/2020-April/094294.html

mlennert:
BTW: I'm starting to have difficulty knowing where to discuss dev related questions. More and more is happening on GitHub, with less going to the mailing list...

neteler:
I agree and would like to cc these conversations to the grass-web mailing list, any ideas?

veroandreo:
ping @nbozon  :)

can we merge this? It makes it easier to build the site locally and test

hellik:
> 
> 
> I agree and would like to cc these conversations to the grass-web mailing list, any ideas?

I concur here too

neteler:
> I agree and would like to cc these conversations to the grass-web mailing list, any ideas?

I have now modified the repo settings:

> Notifications 
> > Setup email addresses to receive notifications when push events are triggered.
> > > Address
> > > > grass-commit@lists.osgeo.org grass-web@lists.osgeo.org

This entry here will show if it works.

PR #76, opened 2020-04-03
sample data: add image, unix line endings

### LINKED ISSUE #71: Sample data: illustrate download from GRASS with a screenshot
Implement Moritzs suggestion to iIllustrate downoad from GRASS GIS with a screenshot like [this one.](https://trac.osgeo.org/grass/attachment/wiki/Grass7/NewFeatures74/grass740_startup_sample_data_download_arrow.png)

Maybe wait until gallery content is uploaded?

ninsbl:
Fix #71

### DIFF
--- a/content/download/data.en.md
+++ b/content/download/data.en.md
@@ -13,12 +13,13 @@ layout: "data"
 [ [**North Carolina dataset**](#NorthCarolinaDataset) | [**Spearfish dataset**](#SpearFishDataset) | [**Datasets for special topics**](#SpecialTopicDatasets) | [**Global datasets**](#GlobalDatasets) ]
 
 <div class="alert rounded-0 alert-default">
-<i class="fa fa-arrow-right"></i> This section gathers sample datasets available as ready-to-use GRASS GIS <b>locations</b> and sample data in other common GIS formats.
+<i class="fa fa-arrow-right"></i> This section gathers sample datasets for learning or testing GRASS GIS. Datasets are made available as ready-to-use GRASS GIS <b>locations</b> and sample data in other
+common GIS formats.
 
 </div>
 
 <div class="alert rounded-0 alert-default">
-<i class="fa fa-arrow-right"></i> Note: an easy way to download sample data is from the download button on the GRASS GIS startup screen.
+<i class="fa fa-arrow-right"></i> Note: an easy way to download sample data is from the <a href="#DownloadGUI">download button on the GRASS GIS startup screen</a>.
 </div>
 
 <hr>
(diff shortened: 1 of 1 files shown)


mlennert:
I think we can go ahead and merge. In the current state I think it is better to release early and often to staging.

lucadelu:
@mlennert sorry but I reviewed it and make comments, why did you merge? Are we now in hurry after 4 years of nothing? If you like to test it you can download the diff and test on your machine.

@ninsbl please consider my review since having the images at the end of the page doesn't make any sense

ninsbl:
No worries. Will creare a new PR with the requested changes.

mlennert:
Sorry @lucadelu I don't see your comments. Where are they ?

As mentioned in my message I believe that we will be more efficient if we commit early and often as this is still staging, not a public site. IMHO, having to download diffs throughout the different PRs makes it much more complicated to get a good picture of where we are at. See e.g. Helli's question on the list

But if others don't agree, I'm willing to change course, obviously.

This said, looking at your comment here, I have to admit that I actually liked the proposed solution of mentioning the option of the download button at the top, but giving details (incl the image) at the end.

lucadelu:
> Sorry @lucadelu I don't see your comments. Where are they ?
> 

@mlennert there was a review now seems to be cancelled

> As mentioned in my message I believe that we will be more efficient if we commit early and often as this is still staging, not a public site. IMHO, having to download diffs throughout the different PRs makes it much more complicated to get a good picture of where we are at. See e.g. Helli's question on the list
> 
I don't understand what is the difference to wait a day, usually this is the time to answer to review in this moment, it is like to the main code, I think we should use the same method, if no comment merge otherwise wait

> But if others don't agree, I'm willing to change course, obviously.
> 

I prefer to see a "final" version only instead of several one, also for reviews because they could start to work and also the original author is working on...

> This said, looking at your comment here, I have to admit that I actually liked the proposed solution of mentioning the option of the download button at the top, but giving details (incl the image) at the end.

I think the solution is like in the manual or books, if we you are speaking of something you add images close to the text and not all to the end of the page.
@veroandreo  @neteler @nbozon what do you think?

mlennert:
@lucadelu I didn't see your review when I merged either, otherwise I obviously wouldn't have ignored. As this was just the follow-up of a previous merge and my comment to it, I didn't think that we needed much review.

I'll refrain from rapid merging in the future.

veroandreo:
> > Sorry @lucadelu I don't see your comments. Where are they ?
> 
> @mlennert there was a review now seems to be cancelled
> 
> > As mentioned in my message I believe that we will be more efficient if we commit early and often as this is still staging, not a public site. IMHO, having to download diffs throughout the different PRs makes it much more complicated to get a good picture of where we are at. See e.g. Helli's question on the list
> 
> I don't understand what is the difference to wait a day, usually this is the time to answer to review in this moment, it is like to the main code, I think we should use the same method, if no comment merge otherwise wait
> 
> > But if others don't agree, I'm willing to change course, obviously.
> 
> I prefer to see a "final" version only instead of several one, also for reviews because they could start to work and also the original author is working on...
> 
> > This said, looking at your comment here, I have to admit that I actually liked the proposed solution of mentioning the option of the download button at the top, but giving details (incl the image) at the end.
> 
> I think the solution is like in the manual or books, if we you are speaking of something you add images close to the text and not all to the end of the page.
> @veroandreo @neteler @nbozon what do you think?

In general, I like illustrations placed as close as possible to where the topic they refer to is being mentioned/discussed

Issue #59, opened 2020-03-29
Add "Try GRASS GIS online" button


ninsbl:
@wenzeslaus produced some cool stuff that allows to try GRASS GIS online, with plenty of nice examples:

Would be cool to have it referenced from the new website.

See: 
http://osgeo-org.1560.x6.nabble.com/Try-GRASS-GIS-online-Binder-mybinder-org-td5412744.html
 or 
http://osgeo-org.1560.x6.nabble.com/Try-GRASS-GIS-online-rollApp-rollapp-com-td5425767.html

veroandreo:
+1! It could go in [Learn](/content/learn) landing page (i.e., in one of those squares) and maybe also in the main page (i.e., home) either as a button or as a rectangle such as those in the Download landing page. What do you think, @nbozon ?

wenzeslaus:
See also how Jupyter project does it:

[Index page](https://jupyter.org/) has "Try it in your browser" button next to "Install JupyterLab" (under JupyterLab section).

The [Try Jupyter](https://jupyter.org/try) page has series of examples for different technologies (Classic Notebook, JupyterLab) or interest groups (R, C++)* as boxes with title, image and description. The URL for the page is `https://jupyter.org/try`.

*Related to that, I think we should have there more than one option too. There is even more than mybinder and rollApp.

ninsbl:
Now I had a look at CoCalc tryout:
https://cocalc.com/app?anonymous=x11
Unfortunately, for me responsetime was quite slow. So I am in doubt if this is good "advertisement".
How does it work for others. Another limitation is that sample data is missing and the free CoCalc instance cannot download…
Any thought? Add CoCalc or leave it?

neteler:
I'd include CoCalc for now, we may add a (grey font) note that the service might be slow.

veroandreo:
I'm not sure about it... They have GRASS 7.4 (old-ish), it is really slow and you have to download a location which at least here did not work... I would not include it for now

ninsbl:
OK, I am currently implementing Luca`s suggestion with a sub-page under learn. Where I for starters add different option (also the less optimal ones).
But now I am wondering, is there anywhere an actinia test instance available that could be added as "Try GRASS GIS as a Service"? And if there is one would that be of interest in this context?

ninsbl:
See also updates in #80

veroandreo:
> See also updates in #80

[...]

PR #80, opened 2020-04-04
Add try online with mybinder

ninsbl:
See: #59 
just a suggestion for how it could be implemented.
Would be nice if text could be on top of the images and a hover-effect could "warn" that l launching binder would take a few seconds...


lucadelu:
I like the idea, but I would keep just a box for `try online` and maybe add a page with some description about the services (for example that it take a while to start, ecc ecc)

ninsbl:
Fully agreed. Will change that accordingly.

themes/grass/layouts/learn/list.html
```diff
@@ -78,12 +78,12 @@ <h2 class="section-title">Get to know the GRASS GIS ecosystem</h2>
  <a href="#" class="px-4 py-4 d-block bg-white htl"><h3 class="mt-2 mb-4"> Coming soon</h3><br><br><br></a>
   </div>
 
-  <div class="col-lg-4 col-sm-6 mb-4">
- <a href="#" class="px-4 py-4 d-block bg-white htl"><h3 class="mt-2 mb-4"> Coming soon</h3><br><br><br></a>
+  <div class="col-lg-4 col-sm-6 mb-4 text-center">
+ <a href="./learn/tryonline" class="px-4 py-4 d-block bg-white htl"><h3 class="mt-2 mb-4"> Try online </h3><img src="../images/www.svg" width="50%" alt=""  align="middle"/></a>
```
lucadelu, line 82: this seems wrong, try with
```suggestion
 <a href="./tryonline" class="px-4 py-4 d-block bg-white htl"><h3 class="mt-2 mb-4"> Try online </h3><img src="../images/www.svg" width="50%" alt=""  align="middle"/></a>
```

PR #91, opened 2020-04-08
WIP: fixing and homogenenizing style in detailed releases list

veroandreo:
I started working on fixing the list of releases (fixed date format, added GIS where missing, change word order, added some bold face here and there), but there's still quite some work to be done. Here a list:
- RC's after 7.2.0 are missing while all other RC are there
- 7.8.1 and 7.8.2 are missing (if this will be a full list, maybe we should also add them here)
- Most of version 6 and earlier releases are not in bold (we could agree here what to put in bold to make it consistent through out the list, I'd suggest only major releases)
- I did not find the figure of the roadmap
- We could split the list in sections, either decades as in History or some other criteria
- Was it called GRASS GIS before version 4?
- 1990 (Winter) GRASS 3.2 released through Internet <-- will links like this one be maintained?

### DIFF
--- a/content/about/history.md
+++ b/content/about/history.md
@@ -98,7 +98,7 @@ On Dec 9, 2007 GRASS GIS started to use
 and [Trac](https://trac.osgeo.org/grass) bug tracker hosted by 
 OSGeo. Other GRASS GIS project infrastructure like the website and mailing 
 lists were (and still are) hosted by OSGeo. Just for fun have a look
-at the <a href="/about/history/web-evolution">GRASS GIS Website evolution</a>.
+at the [GRASS GIS Website evolution](/about/history/web-evolution) page.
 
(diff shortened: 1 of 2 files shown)


neteler:
After merge of #92 there will be the roadmap graphics (https://github.com/OSGeo/grass-website/blob/9967f933aeac77106b79a2cb4954e212386b9d6c/static/images/gallery/development/grass_5_6_roadmap_graphic.jpg) accessible (the original Libreoffice file I am still searching).

Additional idea: this release page could be "beautified" by adding numerous of the historic screenshots from the gallery: https://github.com/OSGeo/grass-website/tree/master/static/images/gallery/gui/ and https://github.com/OSGeo/grass-website/tree/master/static/images/gallery/3D/

(after merge I am happy to go through and add some more historical contextual notes)

Issue #105, opened 2020-04-13
Mobile usage: GRASS GIS SVG logo too small


neteler:
When visiting https://staging.grass.osgeo.org/download/ on a mobile device, the GRASS GIS SVG logo (used for "Addons" and "source code") is shown too small:

![image](https://user-images.githubusercontent.com/1295172/79112013-eaafdf80-7d7d-11ea-837e-769e645357f5.png)

The reason will be the "31%" scaling, see:

https://github.com/OSGeo/grass-website/blob/02a719c4a62e68e483bea6eed1ec2a0191412524/themes/grass/layouts/download/list.html#L70

and

https://github.com/OSGeo/grass-website/blob/02a719c4a62e68e483bea6eed1ec2a0191412524/themes/grass/layouts/download/list.html#L91

Is any other solution possible here?

lucadelu:
I found a solution using object instead img and using css to set the width, @nbozon do you know a better solution?

lucadelu:
implementation in #113

nbozon:
I confirm using an object with width set by media queries does the trick here. (thx @lucadelu , i increased the .grasslogo to 80% ) , display nicely on mobile now, see on staging :
<img width="211" alt="Capture d’écran 2020-05-24 à 10 27 34" src="https://user-images.githubusercontent.com/7135374/82749420-67849f00-9da9-11ea-8dfa-54dbf9ac37ef.png">
This issue can probably be closed @neteler

neteler:
Looks good now, thanks!

PR #113, opened 2020-04-13
working on #105 GRASS GIS SVG logo too small

lucadelu:
this fix the problem with smartphone but it is not working on tablets, @nbozon could you check it?


neteler:
May this PR be merged?

Issue #114, opened 2020-04-14
PR #72 removed all images used in development page


veroandreo:
How can we restore them now? Needed images are:

- bugreport.svg
- document.svg
- translation.svg
- addons.svg
- core_devel.svg

@neteler is it possible to restore them and then fix the path in content/contribute/development?

nbozon:
Or you may replace them using font-awesome icons as follow

Bug:

       <i class="fa fa-bug fa-3x"></i>

Document:

       <i class="fa fa-file-text fa-3x"></i>

Translattion:

       <i class="fa fa-language fa-3x"></i>

Addons (?) :
       
      <i class="fa fa-plug fa-3x"></i>  

Core dev:
       
      <i class="fa fa-gear fa-3x"></i>

lucadelu:
> How can we restore them now? Needed images are:
> 
>     * bugreport.svg
> 
>     * document.svg
> 
>     * translation.svg
> 
>     * addons.svg
> 
>     * core_devel.svg
> 
> 
> @neteler is it possible to restore them and then fix the path in content/contribute/development?

I have them, but I also like the @nbozon idea.. what do you think?

neteler:
I have them locally, too (sorry for the removal by mistake!). I also like the idea of using font-awesome icons, shall we give it a try?

veroandreo:
Yes, let's use @nbozon's suggestion, I like it :)

neteler:
The solution proposed above by @nbozon has been implemented and now looks like this:

https://staging.grass.osgeo.org/contribute/

Question: shall we upload (as per #119, but with path removal modifications) the SVG files anyway?

neteler:
Can we close this or are the SVGs still desired? (AFAIK they are stored in the git history anyway)

Issue #112, opened 2020-04-13
Some HTML layouts and tags are not displayed after updating hugo in the server 


veroandreo:
See for example: 
- https://staging.grass.osgeo.org/learn/overview/ --> only titles displayed, no content nor images
- https://staging.grass.osgeo.org/download/addons/ --> no image
- https://staging.grass.osgeo.org/contribute/development/ --> no images
- https://staging.grass.osgeo.org/about/history/releases/ --> Only section titles displayed, no <li>, <a> or <img>
- https://staging.grass.osgeo.org/about/community/ --> no font awesome icons

When I build locally with hugo-extended v0.69dev installed from source, I get it mostly right except for some issues with img paths and appearance of home. See images below:

![image](https://user-images.githubusercontent.com/20075188/79166511-9ccec200-7de5-11ea-989d-2cc264202c3a.png)

![image](https://user-images.githubusercontent.com/20075188/79166255-0ac6b980-7de5-11ea-8117-c3a067b439eb.png)

![image](https://user-images.githubusercontent.com/20075188/79166300-203be380-7de5-11ea-9f11-d834bac918e5.png)

lucadelu:
@veroandreo I think the problem is with path to images, I think the images were moved in sub directories, If the problem is this please check all the pages :-/

veroandreo:
@lucadelu: In my local build, yes, but in the server the issue is different.

nbozon:
I noticed most initial theme's images were removed from the root of static/images/ after subdirectories were created...  This is not a path problem, but simply missing images. They were deleted [11 days ago](https://github.com/OSGeo/grass-website/commit/e1070d4f68e08c74155d8532fe18245f31977cb2) it seems.. @veroandreo They are still rendered on the server because they are kept in /public from previous builds.

lucadelu:
I still have them in a public folder, I could add back them...

veroandreo:
Even with the fixes to the img file paths (#115), the site still doesn't display some HTML layouts and tags in the server. Meanwhile, it builds fine locally. There's some other problem, @nbozon. Please, have a look at the links I posted originally.

nbozon:
Yes, i also noticed that every (html) bootstrap element is not rendered on "subpages" only.  Everything's rendered locally, but have no clue of what is happening on the server yet. I wonder if this may be related to the new use of [blackfriday](https://github.com/russross/blackfriday).

veroandreo:
[...]

PR #100, opened 2020-04-12
fixes download/addons and donation pages + layouts

veroandreo:
I created a specific layout for AddOns to avoid getting the "GRASS GIS for XX" title that makes sense for different OS but not for addons. 
The PR also fixes img sizing (I changed MD style to HTML tags to be able to control size).  
And I also fixed the logos' path for download landing page.

PR #115, opened 2020-04-14
fix all images paths

veroandreo:
This PR fixes all images paths and now the site builds fine locally with Hugo-extended v0.69-dev (see problems reported in #112). Moreover, it re-fixes `{{.Site.BaseURL}}/images/imagefile` in several html layouts (I had fixed many in #100 but were reverted apparently). I also moved banner.jpg to `static/images/logos/` otherwise it was not discoverable under `themes/grass/images` because `{{.Site.BaseURL}}` points elsewhere.

Furthermore, see #114. All images for development page were removed in #72. How come the site was still looking fine if images were not there any more??


neteler:
> This is ok. Images used to be at the root of /static/images though ... 

Sorry, my bad. For an alternative idea with font-awesome icons, see #114 

> They are still rendered on the server because they are kept in /public from previous builds.

This should not happen since (at time) the entire target directory is cleaned before new deploying. I guess it is a leftover of d39978da86dcb5cdcec7603602f4847304924dbc (added to .gitignore but not removed before?). I am just guessing...

veroandreo:
I merge this one then. Later, I tonight I can implement @nbozon's suggestion for the missing images.

PR #119, opened 2020-04-15
working on #114 (SVG file upload)

lucadelu:


neteler:
Maybe not to merge...

lucadelu:
Closing this, if we want to upload images (but if we don't use I don't see the need to keep them) we can do it in a new PR

PR #131, opened 2020-05-03
Adds search engine

nbozon:
Adds client-side static search functionality as requested -

**Basic concept**
- Hugo will create a JSON file at the root of the site (/public/search/index.json) that will be used as a database (all pages title, url, content).
- When a user visits one page, this file is downloaded to its browser.
- When words are written to the text box, some Javascript code will search for the words in the variables content and get the corresponding URLs.

**Implementation**
- Create simple search page, hidden from section and menu ( /content/ search.md )
- Create basic search template (theme/grass/layouts/_default/search.json)
- Calls vue and axios + custom javascritpt from the footer template (theme/grass/layouts/partials/footer.md)
- Add magnifier icon + expandable search input to the right side of to the breadcrumb (shown on every page)
- Fires a collapsible submenu with results founds displayed as page links

**Notes**
- I am no search expert but this provides the requested feature, for searching the static content we have 
- The search template may be enhanced for more advanced/detailed results if needed
- Website content may be written differently if we want more accurate results (especially titles, subtitles but also how it is ordered in a single page...)

Please try it and review @neteler @lucadelu @veroandreo


veroandreo:
It took me a while to find the search thingy ;)

Where does it search? I tested with a word present in the current page and it is not found

search

I'd really like to have the search functionality in the home page too. As the very last item in the menu, for example, just the icon of the magnifying glass so it does not occupy too much space. Is it possible?

lucadelu:
> Adds client-side static search functionality as requested -
> 

thanks a lot @nbozon 

> 
>     * The search template may be enhanced for more advanced/detailed results if needed
> 
>     * Website content may be written differently if we want more accurate results (especially titles, subtitles but also how it is ordered in a single page...)
> 

could you please improve a little bit this sentence and what could be done to fix these two point, index.json seems not to index the main pages of section (download, learn, about, contribute), @veroandreo probably due this problem you cannot find `time series`

mlennert:
@nbozon any comments on this ? I've just tried it and:

- I think the search tool has also to be visible on the home page. Maybe it would be better to integrate it into the menu line at the very top, not the breadcrumb ?
- Using this seriously slows down page loading times for me.
- I also cannot get any search results. For example, checking in the json file, I see that there are many occurrences of the word vector, but when I type 'vector' nothing happens. Maybe there is an issue with the display of results more than with searching ?

veroandreo:
Search added in #334, closing

PR #133, opened 2020-05-03
Adds Gallery of screenshots

nbozon:
Adds a screenshot gallery as requested, using images available in static/images/gallery/*/

Once the site is builded against this branch, please head to /learn/gallery  (or use the learn menu) to see the results

@neteler @ninsbl @lucadelu  @veroandreo  please review


veroandreo:
Thanks, Nick

In my opinion, it is a bit long though, i.e., a lot of browsing. Is it possible to get only one or two lines of screenshots per topic and a `See more` or just `More` link either above or below the pics?? See below a screenshoot of what I try to say:

image

On the other hand, I had also imagined a gallery in the home page, below the `News` block. Something like a single row of 2 or 3 sliding screenshots. Is that possible?

lucadelu:
> Thanks, Nick
> 
> In my opinion, it is a bit long though, i.e., a lot of browsing. Is it possible to get only one or two lines of screenshots per topic and a `See more` or just `More` link either above or below the pics?? See below a screenshoot of what I try to say:
> 

I also think this should be good, @nbozon we could set `height:50vh` for each `div gallery` and add a button `see more` in the bottom of the div to remove the height in the css and enlarge the gallery 

> image
> 
> On the other hand, I had also imagined a gallery in the home page, below the `News` block. Something like a single row of 2 or 3 sliding screenshots. Is that possible?

This is different topic, however it should be possible adding some javascript code to homepage

mlennert:
@nbozon could you let us know your reaction to this ?

mlennert:
I agree with the need to reduce the number of images on the "landing page" of the gallery. This will also allow to reduce the bandwidth needed to load this page.
However, maybe a first, easy solution would be to reduce the number of images in the first place (possibly moving some to a reserve directory). Currently there is tremendous redundancy in these screenshots (I count 10 very similar network analysis images, for example). Maybe some selection in those would be helpful. We could possibly say that we should only have max 4-5 images per theme on this page, with a see more button linking to sub-pages with the complete collection per theme ?
This said, I actually think that a screenshot gallery with too much redundancy is counter-productive and so we should probably just delete some the images...

lucadelu:
I agree with @mlennert to remove some of the images with redundancy. The problem is how to choose which images remove; some could be patch together, like the network analysis one to show the different method

mlennert:
I would suggest that at this stage it is a doocracy decision: whoever does it gets to decide ;-)

As long as all the images are kept safely somewhere to not lose the option to get back to them if so desired, I don't think there is a problem.

I like your idea of composite screenshots, but honestly I would really like this website to go public and so I would privilege acceptable paths of least resistance. Just chosing 3-5 representative images per theme would probably be the fastest. Later, we can always change the images later, and build a more sophisticated system with a click bringing you to more screenshots of the same theme.

neteler:
> I would suggest that at this stage it is a doocracy decision: whoever does it gets to decide ;-)

I'd agree to that.

> As long as all the images are kept safely somewhere to not lose the option to get back to them if so desired, I don't think there is a problem.

Also agreed. We already have 
https://github.com/OSGeo/grass-website/tree/master/static/images/gallery/history

but it could also be nearby in a new parallel subdirectory with a different name.

> I like your idea of composite screenshots, but honestly I would really like this website to go public and so I would privilege acceptable paths of least resistance. Just chosing 3-5 representative images per theme would probably be the fastest. Later, we can always change the images later, and build a more sophisticated system with a click bringing you to more screenshots of the same theme.

Yes, please let's go live with a selected set of screenshots. No delaying ... :-) we are in year 4 of getting it out!

neteler:
Gentle ping :-)

neteler:
How to test this?

```
$ wget https://github.com/OSGeo/grass-website/pull/133.diff
$ git apply 133.diff
133.diff:16: trailing whitespace.
	  
133.diff:92: trailing whitespace.
    padding-bottom: 100%; 
133.diff:114: trailing whitespace.
.gallery.hover-effect-zoom .img, 
133.diff:124: trailing whitespace.
    position:relative; 
133.diff:214: trailing whitespace.
    margin: auto; 
error: cannot apply binary patch to 'themes/grass/static/plugins/gallery/default-skin.png' without full index line
error: themes/grass/static/plugins/gallery/default-skin.png: patch does not apply
error: cannot apply binary patch to 'themes/grass/static/plugins/gallery/preloader.gif' without full index line
error: themes/grass/static/plugins/gallery/preloader.gif: patch does not apply
```

Now what?

neteler:
@nbozon @mlennert I have cleaned up the gallery and moved many oldish screenshots out of their respective folders into a new, but `museum/` folder so that nothing is lost.

Time to merge this PR and see how we like it?

mlennert:
> @nbozon @mlennert I have cleaned up the gallery and moved many oldish screenshots out of their respective folders into a new, but `museum/` folder so that nothing is lost.
> 
> Time to merge this PR and see how we like it?

I can't find the screenshot gallery in the website anymore. Could someone remind me how to reach it ?

neteler:
https://github.com/OSGeo/grass-website/tree/master/static/images/gallery

mlennert:
> https://github.com/OSGeo/grass-website/tree/master/static/images/gallery

I meant in the website... ;-)

mlennert:
Looking through the vector directory, there are still a large number of fairly redundant network analysis screenshots. For example, I don't think we need a distance and a time version for each analysis. Obviously it would be nice to have some more screenshots of vector analyses other than network, but someone would have to create those...

neteler:
https://staging.grass.osgeo.org/images/gallery/

mlennert:
> https://staging.grass.osgeo.org/images/gallery/

That gives me a directory listing, not an image gallery. Is that only for me ?

neteler:
There is no gallery unless this PR has been merged... am I wrong?

mlennert:
> There is no gallery unless this PR has been merged... am I wrong?

Duh, obviously. So, after 'git checkout grass-website-gallery' and 'hugo server', I can see it :-)
This is already a bit better, but the original problem of too many images on the landing page is still there. However, I wouldn't consider this a blocker, especially since the menu allows to quickly jump to the desired theme. So, yes, I'm for merging at this stage, with possible future improvements to come in the future.

neteler:
> Looking through the vector directory, there are still a large number of fairly redundant network analysis screenshots. 

Found the problem! Before building the server (via cronjob) I have to delete the previously created `public/` folder in the "hugo space"... I updated the every-15min-cronjob accordingly.

Now the pictures which I moved yesterday to `museum/` are also cleaned up on the server, e.g.
 https://staging.grass.osgeo.org/images/gallery/vector/

mlennert:
> > Looking through the vector directory, there are still a large number of fairly redundant network analysis screenshots.
> 
> Found the problem! Before building the server (via cronjob) I have to delete the previously created `public/` folder in the "hugo space"... I updated the every-15min-cronjob accordingly.
> 
> Now the pictures which I moved yesterday to `museum/` are also cleaned up on the server, e.g.
> https://staging.grass.osgeo.org/images/gallery/vector/

Actually, the cleanup in the gallery was in the master branch, but not in the branch for this PR, so I had to merge master into the branch to actually see the gallery with the cleanup applied.

I would suggest to merge this for now, as I consider it "releasable". I still would love to see the possibility of just showing 3-4 images per theme and then expand that when the user clicks on a theme, but that is a bonus not a must have for release.
@lucadelu and @veroandreo what do you think ?

lucadelu:
Yes, I think is better to merge and make improvements later

veroandreo:
> I would suggest to merge this for now, as I consider it "releasable". I still would love to see the possibility of just showing 3-4 images per theme and then expand that when the user clicks on a theme, but that is a bonus not a must have for release.
> @lucadelu and @veroandreo what do you think ?

I'm in favor of merging too. @nbozon would you mind having a look and commenting here, please?

nbozon:
yes, +1 for merging too ! we can still make it better later on (css/js), and the gallery shortcode can probably be set to limit the numbers of items we want to display

neteler:
I am also in favour of merging now - otherwise the remaining open issues mentioned above with the gallery are hard to judge.

neteler:
Does anyone dare to merge this PR? :smiley:

veroandreo:
done

PR #134, opened 2020-05-03
clean up download linux page

landam:
* new stable -> current
* comment-out broken stuff (grass-daily)
* move generic links to upper part

Screenshot_2020-05-03 Linux(1)

### DIFF
--- a/content/download/linux.en.md
+++ b/content/download/linux.en.md
@@ -8,56 +8,54 @@ layout: "os"
 
 #### Quick links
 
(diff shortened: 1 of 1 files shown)


veroandreo:
Personally, I prefer how it was before, with the focus on the grass versions and Linux distros download links under the respective versions. This was partially discussed in the call we had some weeks ago. However, let's see what others say

Great that you removed the broken stuff and fixed links under 76 that were pointing to 78. Good catch :)

landam:
>  it was before, with the focus on the grass versions and Linux distros download links under the respective versions. This was

@veroandreo But most of links are generic(!) (those I moved on top), eg. https://packages.debian.org/grass and not GRASS-version specific.

mlennert:
I like @landam's proposal. I think that for most GNU/Linux users seeing that the solution linked to their distribution is the most important.

Issue #70, opened 2020-04-01
Improve content of First Time User page


mlennert:
Ideas:

- pictures (but we have to see about rights): see #67 for some suggestions of images

- add a glossary page with some important GRASS GIS specific concepts (database structure, computational region, vector layers, etc...)

wenzeslaus:
Possible starting point (at least for the list and some definitions):

https://grasswiki.osgeo.org/wiki/Terminology_comparison_between_ArcGIS_and_GRASS_GIS

veroandreo:
I added some other links and screenshots in #125

lucadelu:
I suggest to move the videocamera icon up to the text just down the title

veroandreo:
> I suggest to move the videocamera icon up to the text just down the title

I agree, done and merged in #136 :)

lucadelu:
Thanks, are you able to add some more words to have one more line in the first sentence?

(for a graphical point of view should be better to have the list at the same space in the left border, we can fix it also by css, however the text solution is good and it work already in table e mobile view, only desktop has the problem)

neteler:
To me the page now looks quite nice:
https://staging.grass.osgeo.org/learn/newcomers/

veroandreo:
> Thanks, are you able to add some more words to have one more line in the first sentence?
> 
> (for a graphical point of view should be better to have the list at the same space in the left border, we can fix it also by css, however the text solution is good and it work already in table e mobile view, only desktop has the problem)

addressed in #145, please check

mlennert:
>     * add a glossary page with some important GRASS GIS specific concepts (database structure, computational region, vector layers, etc...)
After some reflection, I think this would probably be better off on the wiki than on the website. What do you think ?

veroandreo:
> > ```
> > * add a glossary page with some important GRASS GIS specific concepts (database structure, computational region, vector layers, etc...)
> > ```
> 
> After some reflection, I think this would probably be better off on the wiki than on the website. What do you think ?

Since in the `First time user` page we are providing links to the manual pages and some intro tutorials and courses where all these concepts are explained in full, I do not think we need to add a glossary in the website

veroandreo:
Closing, feel free to re-open if fixes are not satisfactory or just make new PRs with enhancements :)
For a first release, I believe this page is fine now

PR #145, opened 2020-05-10
fix for #70

veroandreo:
I just added a couple more sentences to `First time users > Videos` section so the film camera appears surrounded by text as suggested by @lucadelu in https://github.com/OSGeo/grass-website/issues/70#issuecomment-623186769

Looks like this now:
image


mlennert:
I like this, thanks !

Issue #135, opened 2020-05-03
Allow bigger sizes for font awesome icons


veroandreo:
Following the logic of the css in `themes/grass/static/plugins/fontawesome/css/font-awesome.css`, I tried adding larger sizes
```
.fa-7x {
  font-size: 7em;
}
.fa-9x {
  font-size: 9em;
}
```
but it didn't change anything. Indeed, with size > 5x the icon gets small again. Is there something else that should be changed to make it work?

lucadelu:
I don't know why but .fa-7x and .fa-9x are missing in font-awesome.css, added in our style with #151

PR #151, opened 2020-05-20
working on #135

lucadelu:


veroandreo:
Does it look ok? When I tried the exact same changes in css locally, the icons looked smaller with fa-7x and fa-9x... that's why I created the issue.

lucadelu:
yes here you are an example
icons

veroandreo:
Great! Please merge :)

Issue #120, opened 2020-04-19
What to add in the side panel of 'Get involved' landing page?


veroandreo:
What shall we add in the side green panel of the `Get involved` landing page?? Links to what? Pictures? 

@nbozon, any ideas??

lucadelu:
I would like to remove it, most of the pages have just one column, I would like to remove the side panel also from the others pages having it (About -> community, About -> Credits)

veroandreo:
I don't want to remove them, I like them indeed. Otherwise pages are too boring. But, I'd like to hear some ideas of what to put there or which other layout to use.

veroandreo:
@nbozon how do I edit the `themes/grass/layouts/contribute/contribute.html` file in order to see sections' links in the side-panel?

lucadelu:
I was checking again the website and I found again really bed two see two columns in `Contribute` page. Please look in order `Download`, `Learn`, `Contribute`, `About` pages and let me know if you don't see something strange in `Contribute`, all the others pages are one column only (`Learn` is not completely one column but is one row)

mlennert:
I'm more with @lucadelu on this one. I don't really like those green side panels either, especially since they significantly reduce the width of the main body, but also because I do think having a homogeneous look and feel to all the pages is actually better (even if boring...).

ninsbl:
On this one I also agree with @lucadelu. Esp. if we have challenges with filling them with useful information. Note also that they behave a bit strange on mobile (stay on top of the page).
At least for the time being I would just remove them to get the cool new website online. They can always be added back once we have a good plan for how to use them...

veroandreo:
Fine for me, please go ahead... Still, it would be nice to understand how to edit them if we want to add them back in the future

PR #152, opened 2020-05-21
Sidepanel

lucadelu:
Removed side panel for Contribute landing page, see #120


veroandreo:
would you mind including a screenshot for an easier and quick visualization?

content/contribute/_index.en.md
```diff
@@ -1,15 +1,17 @@
 ---
-title: "Get involved"
+title: "Contribute"
```
veroandreo, line 2: Why did you change this title again? Now the page will have the same title twice, once in the header and again as main title... This was the reason I originally modified it to `Get involved`

lucadelu: @veroandreo sorry but I don't understand, now (with last changes) the page look like this 
contribute

It has the same structure of other landing pages

lucadelu:
> would you mind including a screenshot for an easier and quick visualization?

Done, maybe later we can find a more appropriate image if you don't like this

neteler:
I think like this it looks fine and the issue pointed out by @veroandreo is indeed not re-appearing.

veroandreo:
Looks nice! Please, merge :)

PR #157, opened 2020-05-21
Use unified width for images in book tables

wenzeslaus:
The size of images on learn/books page is driven by the image size instead of the layout. I'm using `max-width: 10vw;` to unify (limit) their widths.

For now I added `max-width` to general `table img`, but that does not seem like a good solution. Please comment.

### DIFF
--- a/themes/grass/assets/css/style.css
+++ b/themes/grass/assets/css/style.css
@@ -296,6 +296,7 @@ table tr td {
 }
 table img {
   box-shadow: 0 5px 15px rgba(0, 0, 0, 0.07) !important;
+  max-width: 10vw;
 }
 .github {
   background: var(--grass-color-light);


lucadelu:
Probably I would like to add a class for images width max-width, because if we have images in other tables with different width this could be a problem.
Since it is markdown we need to have some [hack](https://stackoverflow.com/questions/14732028/add-id-or-class-to-markdown-element) to add class to the table

neteler:
Any ideas? At time the page looks a bit weird:

https://grass.osgeo.org/learn/books/

image

nbozon:
This is now fixed in master, please see https://grass.osgeo.org/learn/books/
This means a new book image can be added and will be correctly resized whatever its original size is.

Done by extending @wenzeslaus' proposal as follows:

- Now using the "books" layout from content/learn/books.md , where the #books id is added on parent container
- Adds specific width, min-width and max-width for the #books table column and img only
- Also adapts the table & img properties for smallest screen (media-query) 

Details can be seen in this [commit](https://github.com/OSGeo/grass-website/commit/059061a5c3ef9f762faf514da6f4f425e811954b).

@wenzeslaus @neteler please check and close this PR if satisfied - Thanks

veroandreo:
Looks fine :+1: Thanks @nbozon!

neteler:
Thanks, 059061a5c3ef9f762faf514da6f4f425e811954b solved it nicely.

PR #161, opened 2020-05-25
added title to news page to have same layout of other landing pages

lucadelu:


lucadelu:
> +1, good idea. The text may be 'News form **the** GRASS GIS world'

done

PR #168, opened 2020-06-06
Gallery cleanup: moved old images into 'museum' folder

neteler:
addresses #133 (gallery overload)


neteler:
Merging since nothing is lost, just moved

PR #64, opened 2020-03-30
added events, #TODO fix allnews.html filter

lucadelu:

Issue #54, opened 2020-03-29
Add a "Next events" section


veroandreo:
It would be nice to have a small section where to put logos (one or two) of next events either in the landing page or in [News](https://staging.grass.osgeo.org/news/). In the latter, the space for the tweets could be shrunk a bit, and add next events logo(s) with link(s) there.

lucadelu:
Started in #64

lucadelu:
"Next events" section is now in News page, in the pull request there was discussion about the position of the section. I have the same idea of @mlennert that it is more useful to have first the Twitter and than "Next events", however is simple to change if majority prefer other solution.

PR #170, opened 2020-06-07
added events to news page

lucadelu:
new pull request for #64


veroandreo:
@lucadelu would you attach a screenshot, please? It's then much easier to visualize changes and give feedback

lucadelu:
next_events

veroandreo:
Thanks for this @lucadelu :)

Is it possible to add a lil padding on the left of "Next events" title? It appears to me as too close to the border of the box... maybe it's just my perception dunno

And, it occurs to me now, that since "Next events" will at most contain 2 items, we could put it above the twitter feed so we make sure that visitors will see it... and the twitter feed will still be visible too... what do you say @neteler  @mlennert ?

lucadelu:
> Thanks for this @lucadelu :)
> 
> Is it possible to add a lil padding on the left of "Next events" title? It appears to me as too close to the border of the box... maybe it's just my perception dunno
> 

ok, so I suggest to move also @grassgis twitter title

> And, it occurs to me now, that since "Next events" will at most contain 2 items, we could put it above the twitter feed so we make sure that visitors will see it... and the twitter feed will still be visible too... what do you say @neteler @mlennert ?

I will display 3 items, now there are two only because I added only two events after today

mlennert:
> Thanks for this @lucadelu :)
> 
> Is it possible to add a lil padding on the left of "Next events" title? It appears to me as too close to the border of the box... maybe it's just my perception dunno
> 
> And, it occurs to me now, that since "Next events" will at most contain 2 items, we could put it above the twitter feed so we make sure that visitors will see it... and the twitter feed will still be visible too... what do you say @neteler  @mlennert ?
>  
> 

I think it really depends on where we think we will be more active: Twitter or next events. I do have the feeling that in average we have more on Twitter, and more directly linked to GRASS itself, while events are mostly broader, so with that in mind I think the order is good as is.

BTW, we should take out the foss4g event announcements before taking the website public as they are cancelled.

lucadelu:
Going to merge, in case continue to discuss about position in #54

PR #188, opened 2020-07-02
avoid absolute URLs to server

neteler:
There should be no more need to have the URLs absolute.

Yet broken links after merging this PR might need a server-side intervention (e.g. missing redirects).


neteler:
I'll merge to not miss the current "review train" :-)

PR #207, opened 2020-07-28
add page with mirrors sites and link on home page

mlennert:
First attempt at adding a list of mirror sites.
Current list certainly not up to date, so needs to be corrected before release.

I added the link only on the home page. Not sure if it should go somewhere else.

The actual page is in content/about/mirrors as we have no content directly in content/ and so I didn't want to put it there. About seemed the best for now.

PR #211, opened 2020-07-28
fix error in footer layout in #207

mlennert:


veroandreo:
oh, I made a PR for the same fix :) That's why I was getting a merge conflict in #212 

I close that one now

Issue #190, opened 2020-07-03
Download source code


micha-silver:
IMHO. there should be an entry on the Download dropdown menu to the "GRASS GIS source code" link.

veroandreo:
See also #195

nbozon:
There are already too many submenu entries,  and we should avoid to add more at this stage.
And source code is visible enough imo:
A link to source code is already available in the footer of every page (https://github.com/OSGeo/grass).
A dedicated subsection is also linking to source code, at the bottom of the [Download page](https://staging.grass.osgeo.org/download/) itself.
Closing for now, please reopen if necessary

mlennert:
I don't agree with this reasoning. For me, easily locatable source code is a must for a GPL project. We should at least have the link to the github clearly marked. Could one entry to the 'Download' menu be a direct link to the repository ?

neteler:
We may simplify the download menu into, e.g.,

- GRASS GIS installers
- GRASS GIS source code
- Docker images
- Addons
- Sample data

veroandreo:
> I don't agree with this reasoning. For me, easily locatable source code is a must for a GPL project. We should at least have the link to the github clearly marked. Could one entry to the 'Download' menu be a direct link to the repository ?

I think this is, at this point, the easiest solution, just one more entry (*Source code*) in the Download menu that points directly to the github repo of grass (we don't need to create a new page). @neteler's suggestion is nice but it would imply a re-design/merge of linux, windows and mac pages

neteler:
Suggestion: move "source code" entry on page https://staging.grass.osgeo.org/download/ to the top, rather than hard-to-reach bottom...

lucadelu:
> > I don't agree with this reasoning. For me, easily locatable source code is a must for a GPL project. We should at least have the link to the github clearly marked. Could one entry to the 'Download' menu be a direct link to the repository ?
> 
> I think this is, at this point, the easiest solution, just one more entry (_Source code_) in the Download menu that points directly to the github repo of grass (we don't need to create a new page). @neteler's suggestion is nice but it would imply a re-design/merge of linux, windows and mac pages

Probably could be better to link the [wiki page](https://grasswiki.osgeo.org/wiki/Compile_and_Installl)

lucadelu:
> Suggestion: move "source code" entry on page https://staging.grass.osgeo.org/download/ to the top, rather than hard-to-reach bottom...

I don't know, depending on target.. do we want to attract normal user or power user/developer ?

[...]

PR #215, opened 2020-07-29
move link to source code to the top of download landing page

veroandreo:
Following suggestion in https://github.com/OSGeo/grass-website/issues/190#issuecomment-664411158, this PR moves the link to download the source code to the top of `Download` page to make it more visible while keeping the other download options visible too

image

### DIFF
--- a/themes/grass/layouts/download/list.html
+++ b/themes/grass/layouts/download/list.html
@@ -32,7 +32,21 @@
             </div>
           </div>
         </div>
+
+
         <div class="container">
+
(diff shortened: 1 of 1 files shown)


mlennert:
I'm more on Markus' side of the argument, here. The other download options remain clearly visible below (or we could integrate a ToC in the page - cf screenshot gallery) and I think that a little bit of proactive proselytism and teaching the user about what free software is about is not a bad thing.

veroandreo:
Merging for now to make the source code more visible

Issue #204, opened 2020-07-27
Write a decent 404.html


neteler:
The current 404.html looks ... rather empty:

e.g.
https://grass.osgeo.org/blabla

![image](https://user-images.githubusercontent.com/1295172/88595207-d6f56400-d062-11ea-949b-58c281fb5dbd.png)

May we have a better one please, showing the user where to go? Thanks!

veroandreo:
I just found [here](https://gohugo.io/templates/404/) that HUGO provides a very basic template. We could in the meantime replace the content of `themes/grass/layouts/404.html` with the following

```
{{ define "main"}}
    <main id="main">
      <div>
       <h1 id="title"><a href="{{ "/" | relURL }}">Go Home</a></h1>
      </div>
    </main>
{{ end }}
```
until @nbozon helps us with something fancier :smiley_cat: 

Here's what academic provides for example: https://gitlab.com/veroandreo/veroandreo.gitlab.io/-/blob/master/themes/hugo-academic/layouts/404.html

lucadelu:
> I just found [here](https://gohugo.io/templates/404/) that HUGO provides a very basic template. We could in the meantime replace the content of `themes/grass/layouts/404.html` with the following
> 
> ```
> {{ define "main"}}
>     <main id="main">
>       <div>
>        <h1 id="title"><a href="{{ "/" | relURL }}">Go Home</a></h1>
>       </div>
>     </main>
> {{ end }}
> ```

Did it... but I was not able to test since if I look for a missing page hugo (using hugo server) return "404 page not found"

neteler:
In Apache, we have

```
grep ErrorDocument /etc/apache2/sites-enabled/000-default.conf 
ErrorDocument 404 /404.html
```

If I remove that, then I can the plain
```
Not Found

The requested URL was not found on this server.
```

With ErrorDocument active, it shows the content of https://grass.osgeo.org/404.html

neteler:
@lucadelu found it: you didn't push it to master!

This isn't updated yet:
https://github.com/OSGeo/grass-website/blob/master/themes/grass/layouts/404.html

lucadelu:
> @lucadelu found it: you didn't push it to master!
> 
> This isn't updated yet:
> https://github.com/OSGeo/grass-website/blob/master/themes/grass/layouts/404.html

I was speaking locally on my laptop. I know I didn't push to master (I forgot to open a PR)

neteler:
[...]

PR #217, opened 2020-07-29
clean up 404.html template

mlennert:
This is an attempt at fixing #204 by simply cleaning the 404.html template a bit.


neteler:
Thanks. While @nbozon has a local different version, I'll merge this for now to avoid that users get confused without any error message...

PR #240, opened 2020-12-28
Added images to gallery

baharmon:
Added two images to the gallery.


lucadelu:
The txt files are not needed and they will not be used. If you want to add info to image I think is better to add them directly in the images

neteler:
I would keep them, as it is the only way to have a description and a source (nobody will go and search for the old PR). The few bytes don't harm...

lucadelu (approved):

veroandreo (approved):

Issue #19, opened 2020-02-15
Add search functionality to website


neteler:
It would be great to have website search (at time not implemented at https://staging.grass.osgeo.org/ while the old CMS offers it).

Apparently, there are different technologies available for that:
 
https://gohugo.io/tools/search/

veroandreo:
The academic theme had a [built-in search](https://github.com/gcushen/hugo-academic/blob/master/assets/js/academic-search.js) that might be useful too. It is then displayed as in: https://sourcethemes.com/academic/ (the magnifier glass in the upper right part :))

nbozon:
Basic static "search engine" implemented and proposed in #131
Please test and reviw @neteler @lucadelu @veroandreo

veroandreo:
Hugo-academic was renamed to [wowchemy](https://github.com/wowchemy/wowchemy-hugo-modules) (MIT license). Here are the search functionalities provided:
- built-in: https://github.com/wowchemy/wowchemy-hugo-modules/blob/master/wowchemy/assets/js/wowchemy-search.js
- Algolia: https://github.com/wowchemy/wowchemy-hugo-modules/blob/master/wowchemy/assets/js/algolia-search.js

Maybe we can just adapt one of them accordingly and re-use. I cannot really judge how difficult it might be. 

The built-in one works nicely in my website. However (dunno if relevant for grass website), wowchemy docs say:
> Note that this search engine runs entirely in the web browser on the visitor’s device. Thus, it is not scalable to an extremely large number of pages. If you have a very large amount of content and find the search running slowly, we recommend opting for a server-side search engine such as Algolia or Google

@nbozon, @lucadelu what do you think ?

lucadelu:
> 
> @nbozon, @lucadelu what do you think ?

No clear idea, I looked at the code and this is only the client side there should be something more (like json document with all the info and some html code).
Sorry but I have no time, now and probably in the next months, to spend on this topic

veroandreo:
> > @nbozon, @lucadelu what do you think ?
> 
> No clear idea, I looked at the code and this is only the client side there should be something more (like json document with all the info and some html code).

This is what I found:
- https://github.com/wowchemy/wowchemy-hugo-modules/blob/master/wowchemy/layouts/index.json
- https://github.com/wowchemy/wowchemy-hugo-modules/blob/master/wowchemy/layouts/partials/search.html

> Sorry but I have no time, now and probably in the next months, to spend on this topic

[...]

PR #242, opened 2021-01-01
Search engines: local and DuckDuckgo

neteler:
Awaiting #131 to become ready, here an **interim solution** which addesses #19:
- new page https://grass.osgeo.org/about/search/ with menu entry under "About"
   - Web site search is done locally (the preview looks so-so but it doesn't slow down the site)
   - Manual pages search: via DuckDuckgo unless a better solution is implemented

To show this PR at work I have implemented it on the server.

On mobile devices it looks like this:

image

The code comes from here: [Volltextsuche für HUGO](https://weitblick.org/post/simple-static-site-search-hugo-jamstack/) (sorry, in German language - try to [machine translate it](https://translate.google.com/translate?sl=auto&tl=en&u=https://weitblick.org/post/simple-static-site-search-hugo-jamstack/)).


lucadelu:
I looked at the code but I was not able to test it, I would like to do in the next days.
For manuals we could create during compilation a JSON file like the one done for the website and read info from it.

lucadelu:
I tested it and it seems work properly, there is only a problem. If you switch from AND to OR without any work in the search field it return  `undefined` that it is not beautiful to see. I tried to look on the problem but I was not able to find the solution. I'll try to look on it next day.

neteler:
I have reverted the patch on the server (it was there for a month locally applied to see how well it works) as it breaks our RSS feed (https://grass.osgeo.org/index.xml) which was reported in issue #247.

Hence: this PR needs a change to not break RSS (or, we improve the open #131). In any case, we need a "search" field!

veroandreo:
Search added in https://github.com/OSGeo/grass-website/pull/334, closing

PR #244, opened 2021-01-16
rename versions as suggested in mailing list: current, legacy and preview

veroandreo:


neteler:
Thanks for the PR.

There is the suggestion to just call it "current", see https://lists.osgeo.org/pipermail/grass-dev/2021-January/094841.html

veroandreo:
> Thanks for the PR.
> 
> There is the suggestion to just call it "current", see https://lists.osgeo.org/pipermail/grass-dev/2021-January/094841.html

just fixed

PR #249, opened 2021-01-30
WinGRASS-7.8.5-2-Setup-x86_64 published

landam:
New 64bit wingrass binary available (see https://github.com/OSGeo/grass/issues/1218)


neteler:
Great, thanks. 
Given its importance, should we additionally write an extra news entry?

hellik (approved):

hellik:
> Great, thanks.
> Given its importance, should we additionally write an extra news entry?

yes, would be good

hellik:
And a post in the ML

landam:
No strong opinion here, probably post in the ML could be enough.

hellik:
> 
> 
> No strong opinion here, probably post in the ML could be enough.

posted now in the user ML, see https://lists.osgeo.org/pipermail/grass-user/2021-January/082155.html

PR #267, opened 2021-09-19
add MARWAN new mirror to the list

veroandreo:


lucadelu (approved):
I will wait since they don't fix the internal links

veroandreo:
They just fixed it, and all works now. Merging :)

PR #285, opened 2022-02-14
Download section: replace command-like link by a button

veroandreo:
After adding the extra links in the Source code row in https://grass.osgeo.org/download/, the command-like link to the github repo does not appear when the site is displayed on mobile phones or when zooming in above 170% (in my laptop screen). Hence, I thought of replacing it by a button saying `Clone me` similar to the download buttons below. It looks like this: 

image

and it does not disappear when zooming in a lot. On phones, I do not know. Anyway, an idea. What do you think?

PS: branch name does not really reflect the change as I originally thought it was a problem with row height (it might be indeed), but I didn't find a way to fix that. Hence, the button idea :wink:


cmbarton (approved):
Looks fine on my desktop and phone. Thanks for updating this.

wenzeslaus (approved):

neteler (approved):

PR #293, opened 2022-03-17
Fix wingrass preview link -> pointing at directory

landam:
Currently WinGRASS 8.2 (preview) points at out-dated build:

https://wingrass.fsv.cvut.cz/grass81/WinGRASS-8.1.dev-e3080fd7a-35-Setup.exe

Correct link should point at:

https://wingrass.fsv.cvut.cz/grass81/


veroandreo (approved):
Thanks @landam ! :)

PR #297, opened 2022-03-28
Add MacPorts to Mac download page

nilason:
Now finally, GRASS 8 is available with the `grass` port.
This also adds the "old" port `grass7`.


nilason:
Right,  the "Download" link to official binary (to http://grassmac.wikidot.com/downloads) could be accompanied by a download icon (event though it's not a direct download link as for Windows). But I think the "MacPorts" link is more similar to the Linux distributions.
What about "icon+Download" and "no icon+Macports"?

neteler:
Looks good to me:

image

neteler (approved):

veroandreo (approved):

nilason:
Thanks!

Issue #298, opened 2022-04-16
Main web page lacks intro sentence


neteler:
The main page at https://grass.osgeo.org/ lacks a one-liner to explain what GRASS GIS is offering:

![image](https://user-images.githubusercontent.com/1295172/163688465-d79970de-2199-4a45-8993-2b03742d8333.png)

Compared to this, the **old web site** version offered a (too long) sentence:

![image](https://user-images.githubusercontent.com/1295172/163688537-db833434-f707-48ff-8099-c4bc292495c4.png)

Suggestion: add a short statement under the line

"Geographic Resources Analysis Support System"

to be more inclusive. The short text would also improve our SEO ranking.

veroandreo:
The line below which you'd like to add the description is this one: https://github.com/OSGeo/grass-website/blob/978662a51a9e22272093af1a07709a2852df3905/themes/grass/layouts/index.html#L18

I made a test by adding an extract of what now appears as website description if you search for grass in google... This is how it looks:

![image](https://user-images.githubusercontent.com/20075188/163692856-b6b7202d-a51b-4b0f-b5ff-08bcdf40fda1.png)

Personally, I'm not fond of the result. I'd prefer to keep it simple as there's a big learn button there and a learn menu above. However, let's hear others' opinions.

petrasovaa:
Or let's ditch the Geographic Resources Analysis Support System and just have the info line.

neteler:
> Or let's ditch the Geographic Resources Analysis Support System and just have the info line.

Yes, the expanded acronym could go into "Learn" section and may be replaced with the explanatory text.

For a SEO check, use e.g.:

https://seorch.eu/seo-check.html

veroandreo:
Sounds good to me, I assigned it to myself to make some tests. However, I really appreciate your contribution in terms of a couple of sentences to add there, especially to avoid repeating the same text over and over again :sweat_smile:

neteler:
Here a suggestion:

_GRASS GIS offers powerful raster, vector, and geospatial processing engines in a single integrated software suite. It includes tools for spatial modeling, visualization of raster and vector data, management and analysis of geospatial data, and the processing of satellite and aerial imagery. It comes with a temporal framework for advanced time series processing and a Python API for rapid geospatial programming. GRASS GIS has been optimized for performance and large geospatial data analysis._

wenzeslaus:
[...]

PR #299, opened 2022-04-22
Add GRASS description in main page - addresses #298

veroandreo:
This PR removes the text Geographic Resources Analysis Support System from the main page and replaces it by a general description about GRASS, as suggested in #298 

Before this PR: 
image

With this PR:
image


neteler (approved):
Thanks!

PR #305, opened 2022-05-17
content/download: fix MS Windows standalone installer GRASS GIS 8.3.0 (preview) version download link

tmszi:
Fix [MS Windows standalone installer GRASS GIS 8.3.0 (preview) ](https://grass.osgeo.org/download/windows/)version download link. Additional question: Last version [WinGRASS-8.3.dev-394970d61-92-Setup.exe](https://wingrass.fsv.cvut.cz/grass83/) is from 2022-05-06 17:06, Is something wrong with the build process?

### DIFF
--- a/content/download/windows.en.md
+++ b/content/download/windows.en.md
@@ -44,13 +44,13 @@ Standalone installer: install GRASS GIS with the required support packages.
 
 <hr>
 
-#### <a name="GRASS-GIS-devel"></a> GRASS GIS 8.2.0 (preview)
+#### <a name="GRASS-GIS-devel"></a> GRASS GIS 8.3.0 (preview)
 
 <div class="alert rounded-0 alert-info">
 <i class="fa fa-info-circle"></i> Active <u>development</u> and <u>experimental</u> <b>GRASS GIS</b> version.
 </div>
 
-*  [<i class="fa fa-download"></i> Download 64bit](https://wingrass.fsv.cvut.cz/grass81/)
+*  [<i class="fa fa-download"></i> Download 64bit](https://wingrass.fsv.cvut.cz/grass83/)
 
 <hr>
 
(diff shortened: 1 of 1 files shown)


veroandreo:
This change would require to also update all other download subpages where 8.2 is mentioned (i.e., Linux, Mac, Docker) as well as all section links in the upper part of each subpage... I was waiting for 8.2 final release to apply them. Mainly because if we switch preview to 8.3, it seems inconsistent with the rest as all versions showed are even (7.8, 8.0, 8.2). So, with the current release "policy" we should indeed change it to 8.4 (preview). Also, once 8.2 is out, shall we keep 7.8 there or not? 

In any case, this might change if we adopt the RFC about version numbering in https://github.com/OSGeo/grass/pull/2357 and we start releasing all even and odd numbers for minor versions. Opinions?

echoix:
> Also, once 8.2 is out, shall we keep 7.8 there or not?

I think we should for a bit, since it is the version of GRASS that is bundled with QGIS (osgeo4w), including QGIS 3.24.3 released 4 days ago, and their LTS. Even with the latest QGIS 3.24.3, the processing algorithm does not seem to be designed to use grass8, it uses grass7.

So, having so vast usages of grass 7.8, I think we should at least have the option of downloading 7.8 on the side for a long, similar to a "LTS", since its the last release of the Grass 7 series, in case there is a breaking change with Grass 8.

Maybe bump the old versions with a link to "older installer versions"/"archived downloads" like we see in many websites should be useful?

neteler (approved):
Fine, as soon as G8.2.0 is released.

neteler:
Can this be merged?

veroandreo:
no need to, I believe, as the change suggested is included in #308, AFAIU

neteler:
> no need to, I believe, as the change suggested is included in #308, AFAIU

Compared and right, covered by #308. So I am closing this PR.

PR #308, opened 2022-06-03
News: GRASS GIS 8.2.0 released

wenzeslaus:


neteler:
@wenzeslaus Minor edits and screenshot suggestion as a PR against this PR:
https://github.com/wenzeslaus/grass-website/pull/1

content/news/2022_06_03_GRASS_GIS_8_2_0_released.md
```diff
@@ -0,0 +1,64 @@
...
+author: GRASS Development Team
+---
+
+### What's new in a nutshell
+
+The 8.2.0 release of GRASS GIS is now available with results from the GSoC 2021
+and many other additions. A new _grass.jupyter_ package is now included for
+interacting with Jupyter notebooks. Single window graphical user interface
+is available in settings. _r.series_ and three other modules are newly parallelized.
+Additionally the releases includes a series of scripting, packaging, and reproducibility
```
veroandreo:
```suggestion
Additionally the release includes a series of scripting, packaging, and reproducibility
```

content/news/2022_06_03_GRASS_GIS_8_2_0_released.md
```diff
@@ -0,0 +1,64 @@
...
+improvements.
+
+### Full list of changes and contributors
+
+For all 220+ changes, see our detailed announcement with the full list of
+features and bugs fixed at
+[GitHub / Releases / 8.2.0](https://github.com/OSGeo/grass/releases/tag/8.2.0).
+
+Thank you all 28 contributors including 5 new contributors! See the release
+link above for a list and see
```
veroandreo, line 31:
```suggestion
link above for a list and 
```

veroandreo (commented):
I marked very minor details. This PR covers the fix proposed in #305 for windows only, so we can close that one

content/download/docker.en.md
```diff
@@ -9,7 +9,7 @@ layout: "os"
   
 #### Quick links
 
-[ [**GRASS GIS 8.0 (current)**](#GRASS-GIS-current) | [**GRASS 7.8 (legacy)**](#GRASS-GIS-old) | [**GRASS 8.2 (preview)**](#GRASS-GIS-devel) ]
+[ [**GRASS GIS 8.2 (current)**](#GRASS-GIS-current) | [**GRASS 7.8 (legacy)**](#GRASS-GIS-old) | [**GRASS 8.3 (preview)**](#GRASS-GIS-devel) ]
```
landam, line 12: Legacy version should be changed to 8.0, I would say.

veroandreo: Originally, I thought the same, but @echoix raised a valid point in https://github.com/OSGeo/grass-website/pull/305#issuecomment-1129356403 regarding why to keep 7.8 for a while longer...

wenzeslaus: On a PSC meeting some time in the past, we have discussed (_not approved_), under [Scenario 2](https://trac.osgeo.org/grass/wiki/Release/Schedule#Scenario2), a policy where we, as project, don't actually support any other releases than the current release:

> We have only one actively updated, maintained, and supported release series.

However, under this scenario, we do support people updating any older release series when they choose to do so:

> These versions (series/branches) are no longer actively updated, but are updated on demand, i.e., if you submit a patch to fix a bug, we will likely accept it and create a new release when there is enough changes accumulated.

Under this scenario, @neteler (based on latest commits and releases) is the person who is maintaining 7.8 as an individual contributor based on his own decision while the project stopped supporting 7.8 already. (This is obviously just good old do-ocracy, but expressed more clearly.) Under this scenario, we are just stopping support of 8.0 with the release of 8.2.0. And because there is no special treatment of the "one before current" release series, 7.8 and 8.0 are on the same level, i.e., not supported, although contributors can successfully push for a next maintenance release of either of those.

Under this scenario, given the above, it is @neteler's decision to keep 7.8 here or switch to 8.0 or, e.g., even support both, but don't have them listed here.

As I said, we didn't vote on this nor made any decision later as far as I recall.

I think this would be also different from just keeping links to old releases ("older installer versions"/"archived downloads") on the page as part of @echoix comment suggests. These all would not be maintained, but simply a couple of the last Windows installers would be linked (without the need for people to figure out the URL themselves).

content/news/2022_06_03_GRASS_GIS_8_2_0_released.md
wenzeslaus: `releases -> release` fixed, thanks.

content/news/2022_06_03_GRASS_GIS_8_2_0_released.md
wenzeslaus: Second "see" removed, thanks.

wenzeslaus:
Most of our news posts are by _GRASS Development Team_ or _Admin_. Let's try to make it personal and acknowledge people, too. I included myself and added "_and the GRASS Development Team_". Feel free to edit the branch and add yourself. (Note: I don't see that the author field would be actually used, but the release news tend to have some sort of signature.)

content/download/docker.en.md
neteler: > Under this scenario, @neteler (based on latest commits and releases) is the person who is maintaining 7.8 as an individual contributor based on his own decision while the project stopped supporting 7.8 already.

Please note that in some Linux distributions G7.8 is included (see https://repology.org/project/grass/versions), e.g. Debian 11 (upcoming Debian 12, then with G8.0), Ubuntu 22.04 (upcoming Ubuntu 22.10, then with G8.0), etc. So "my" do-ocracy effort is to backport important stuff as needed even for G7.8. However, given my limited resources I'd also phase out the G7.8 support in the next months.
In any case, this is a discussion for grass-dev or grass-psc lists rather than in a pull request.

neteler:
Now online at
https://grass.osgeo.org/news/2022_06_03_grass_gis_8_2_0_released/

PR #310, opened 2022-08-01
Hugo update and updated css outline rules

nobeeakon:
- Updated defaultMarkdownHandler as 'blackfriday' was removed from Hugo in [v0.100.2](https://www.opensourceagenda.com/projects/hugo/versions)

- Updated css rules for outline. It is good practice to have the ouline, as it helps locate the current focused element when tab navigating

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -24,7 +24,10 @@ greyColorDark = "#A0A0A0"
   priority = 0.5
 
 [markup]
-  defaultMarkdownHandler = "blackfriday"
+  defaultMarkdownHandler = "goldmark"
+  [markup.goldmark]
+    [markup.goldmark.renderer]
+      unsafe = true
   
 [Languages]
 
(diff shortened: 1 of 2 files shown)


veroandreo:
I haven't tried this locally, but my main concern is will older hugo versions support `goldmark` too?? I think it's important to first understand which hugo version the website is compiled with in the osgeo server where it is hosted. 
@neteler would you mind checking that?

nobeeakon:
definitely, it would be good to have that info somewhere in the README

> I haven't tried this locally, but my main concern is will older hugo versions support `goldmark` too?? I think it's important to first understand which hugo version the website is compiled with in the osgeo server where it is hosted. @neteler would you mind checking that?

neteler:
Current version on the server which has been locally installed:

```
hugo version
Hugo Static Site Generator v0.69.0-4205844B linux/amd64 BuildDate: 2020-04-10T09:12:34Z
```

On Debian, v0.80 seems to be available:
https://packages.debian.org/search?keywords=hugo

but not v0.100.x ...

nobeeakon:
better closing this issue as it is modifying config stuff related with hugo version.

PR #311, opened 2022-08-02
fix: allow outline in buttons and a tags when on focus

nobeeakon:
Allow outline in in buttons and links when on focus. This is good for tab navigation

### DIFF
--- a/themes/grass/assets/css/style.css
+++ b/themes/grass/assets/css/style.css
@@ -164,7 +164,7 @@ a.nws:hover {
   outline: none;
 }
 :focus-visible {
-    outline: 2px solid var(--grass-color-dark) !important;
+    outline: 2px solid var(--grass-color) !important;
 }
 
 .btn {
--- a/themes/grass/assets/css/style.css
+++ b/themes/grass/assets/css/style.css
@@ -164,7 +164,7 @@ a.nws:hover {
   outline: none;
 }
 :focus-visible {
-    outline: 2px solid var(--grass-color) !important;
+    outline: 2px solid var(--grass-color-dark) !important;
 }
 
 .btn {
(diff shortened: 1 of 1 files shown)


neteler:
I applied this locally but what am I supposed to see as a difference? Please add a before/after screenshot, thanks.

veroandreo:
I do not see a different behavior either, what are we supposed to see? :)

veroandreo:
I do not get that neither in Chrome nor in Firefox (maybe because of hugo version? I have 0.83 here), but I am not very fond of how it looks from what I can see in the pic; feels to that it disrupts the aesthetics of the theme.

veroandreo:
Ok, now I see the effect.. it's only when clickin on links... Since links turn lighter when hovering over them, would you mind trying with the light green instead of the darker one ?

nobeeakon:
Hey @veroandreo, yeah, as said in a previous comment  this is actually for when using the  [tabs](https://www.computerhope.com/jargon/t/tabkey-250.jpg) key to navigate the page, not for clicking (fixed that, so now it is not shown when clicking). The idea is that some people use Tabs key and if the outline is disabled you can't really understand which part of the page you are on because there is no visual cue, so it is a bit annoying not to have it. Besides that, it is considered a bad practice to disable it, as it impacts accessibility. Anyways, about the color I modified it to be  --grass-color-dark which is the one that is being used in most cases.

veroandreo:
Oh my.. sorry for slow here! I'm not used to navigate websites with tab that's why I didn't get it before. I've now tested different colors and definitely, the best one to tab-navigate is `--grass-color`, as you used originally. I'd then suggest to change back to that one and then, we merge.  

Thanks for removing the effect for clicking navigation :)

nobeeakon:
@veroandreo  :/ sorry for that, sometimes this things even when simple, are difficult to explain... I have updated the color to use again `--grass-color`

PR #318, opened 2022-08-09
fix/grass version numbers 

### LINKED ISSUE #315: Sync man page versions
In GRASS website:
* download section lists latest version as being 8.2
* manual section lists latest version as being 8.0
* manual itself has a version of 8.0.3dev

Apparently release procedure needs to be updated to correctly update man page links too.

nobeeakon:
## fixes

#315 

## what
Grass version numbers (current, legacy, preview) passed to `/data` folder so it can be a single source of truth and used in multiple places. Using shortcode so it can be used inside markdown files

Updated current version man pages links.


content/learn/manuals.md
```diff
@@ -8,11 +8,11 @@ layout: "manuals"
 
 <ul id="links" class="list-unstyled version">
  <li>
-  <span class="mwl"><a href="/grass80/manuals/index.html " target="_blank"> GRASS GIS 8.0 (current) </a></span>
-  <a href="/grass80/manuals/index.html" class="inl btn btn-primary" target="_blank">View HTML</a> <a href="/grass80/manuals/grass-8.0_html_manual.zip" class="inl btn btn-secondary">Download ZIP</a>
+  <span class="mwl"><a href="/grass82/manuals/index.html " target="_blank"> GRASS GIS {{< grassVersion version="current" type="short">}}  (current) </a></span>
+  <a href="/grass82/manuals/index.html" class="inl btn btn-primary" target="_blank">View HTML</a> <a href="/grass82/manuals/grass-8.2_html_manual.zip" class="inl btn btn-secondary">Download ZIP</a>
```
wenzeslaus, line 12: Can the URLs be driven by the variables, too? `8.2` is already there. `82` can be perhaps another variable. There are two there already, so this would be a third one. (Ultimately, it seems to me that the URLs is what needs to be fixed, i.e., there should be 8.2 as elsewhere.)

wenzeslaus (commented):
Great. This seems to simplify the [release procedure](https://github.com/OSGeo/grass/commits/main/doc/howto_release.md) (which now even misses some of the website update steps).

content/learn/manuals.md
neteler: > (Ultimately, it seems to me that the URLs is what needs to be fixed, i.e., there should be 8.2 as elsewhere.)

While I agree that URLs should be changed to be based on variables as well, I humbly disagree that they should include periods. Myriads of documents refer to GRASS GIS versions without dots, and breaking with the URL style after decades doesn't seem very useful to me.

nobeeakon: @wenzeslaus , I've added the 82 to the variables and used in the links and stuff. Kept them as they were.

wenzeslaus: > > (Ultimately, it seems to me that the URLs is what needs to be fixed, i.e., there should be 8.2 as elsewhere.)
> 
> ...Myriads of documents refer to GRASS GIS versions without dots, and breaking with the URL style after decades doesn't seem very useful to me.

Out of scope for this PR, but would that be actually a problem? If they need to be updated, what is the difference between changing from `82` to `83` and changing from `82` to `8.3`? If they don't need to be updated, keeping the old URL is enough to keep the old thing working. If you would still need URLs from both the new and old system, there is likely no issue mixing both because you need to specify all numbers explicitly anyway like for the variables in this PR.

`10.0` seems more clear than `100`. `8.11` is definitively more clear than `811`. I think using dots everywhere is more predictable overall because it is consistent for all numbers within one version, while leaving out the dots leads to inconsistent numbers within one version (although being consistent across versions). However, clarity is anyway likely more important than consistency.

neteler: > Out of scope for this PR, but would that be actually a problem?

Please, let's move this discussion to `grass-dev` as I would like to hear more voices here. In this repo it is too much hidden.

neteler:
@nobeeakon Would you mind to also add
https://github.com/OSGeo/grass-website/blob/bd918461a79b099e847f23ef530d19632e66ae39/static/files/sitemap.xml#L7
? Thanks.

nobeeakon:
@neteler , had to create a new sitemap.xml file to be able to use the values stored in the data. Not sure how this impacts other stuff or if any additional change would be required for this new xml to be used instead of the other one as it is located in a different place

content/learn/manuals.md
marisn: > Out of scope for this PR, but would that be actually a problem? If they need to be updated, what is the difference between changing from `82` to `83` and changing from `82` to `8.3`?

I'd agree with Vaclav. Although it can be discussed on -dev ML,
1. somebody must make a post there;
2. there must be a short deadline to not delay this PR forever.

marisn (approved):

neteler:
Mhh, the "donate popup" PR now conflicted with this one... @nobeeakon would you mind to rebase this PR?

nobeeakon:
fixed @neteler

content/download/windows.en.md
```diff
@@ -37,14 +37,14 @@ Standalone installer: install GRASS GIS with the required support packages.
...
 
 <div class="alert rounded-0 alert-warning">
-<i class="fa fa-info-circle"></i> <u>Old stable release</u>, see <a href="https://trac.osgeo.org/grass/wiki/Grass7/NewFeatures78">GRASS GIS 7.8 new features</a> and  <a href="https://github.com/OSGeo/grass/releases/tag/7.8.7">GRASS GIS 7.8.7 announcement</a> for more information.
+<i class="fa fa-info-circle"></i> <u>Old stable release</u>, see <a href="https://trac.osgeo.org/grass/wiki/Grass7/NewFeatures{{< legacyVersionNoDots.inline  >}}{{- .Site.Data.grass.legacy_version_nodots -}}{{</legacyVersionNoDots.inline >}}">GRASS GIS {{< grassVersion version="legacy" type="short">}} new features</a> and  <a href="https://github.com/OSGeo/grass/releases/tag/{{< legacyVersion.inline  >}}{{- .Site.Data.grass.legacy_version -}}{{</legacyVersion.inline >}}">GRASS GIS {{< grassVersion version="legacy" >}}  announcement</a> for more information.
 </div>
 
 *  {{< donateDialog isToggle=true >}}  
-<a href="/grass78/binary/mswindows/native/x86_64/WinGRASS-7.8.7-1-Setup-x86_64.exe" target="blank">
+<a href="/grass78/binary/mswindows/native/x86_64/WinGRASS-{{< legacyVersion.inline  />}}-1-Setup-x86_64.exe" target="blank">
```
neteler, line 47:
```suggestion
<a href="/grass{{< legacyVersionNoDots.inline  >}}/binary/mswindows/native/x86_64/WinGRASS-{{< legacyVersion.inline  />}}-1-Setup-x86_64.exe" target="blank">
```

content/download/windows.en.md
```diff
@@ -21,14 +21,14 @@ layout: "os"
...
+#### <span id="GRASS-GIS-current"> GRASS GIS {{< grassVersion version="current" >}} (current)</span>
 
 <div class="alert rounded-0 alert-success">
-<i class="fa fa-info-circle"></i> <u>New stable release</u>, see the <a href="https://github.com/OSGeo/grass/releases/tag/8.2.0">release announcement</a> for more information.</div>
+<i class="fa fa-info-circle"></i> <u>New stable release</u>, see the <a href="https://github.com/OSGeo/grass/releases/tag/{{< currentVersion.inline  >}}{{- .Site.Data.grass.current_version -}}{{</currentVersion.inline >}}">release announcement</a> for more information.</div>
 
  
 *  {{< donateDialog isToggle=true >}}  
-<a href="/grass82/binary/mswindows/native/WinGRASS-8.2.0-1-Setup.exe" target="blank">
+<a href="/grass82/binary/mswindows/native/WinGRASS-{< currentVersion.inline  />}}-1-Setup.exe" target="blank">
```
neteler, line 31:
```suggestion
<a href="/grass{{< currentVersion.inline  >}}/binary/mswindows/native/WinGRASS-{< currentVersion.inline  />}}-1-Setup.exe" target="blank">
```

neteler:
Thanks for the re-sync.

Will then this sitemap still exist (pointing to the "sub-sitemaps")?
https://grass.osgeo.org/sitemap.xml

neteler (approved):
Great work!
Now only `data/grass.json` needs to be updated in future.

neteler:
I suggest to merge this PR!

veroandreo:
> I suggest to merge this PR!

I agree, please go ahead :)

neteler:
@nobeeakon Today I enjoyed updating the GRASS GIS version to 8.2.1 on the website by simply editing `data/grass.json`. A big thank to you!

PR #321, opened 2022-08-15
Download: added Fedora-COPR for G82 packages

neteler:
Add Fedora-COPR repo for F35 package of GRASS GIS 8.2.x.

(after merge this PR needs to be considered in #318)


veroandreo (approved):
Thanks :)

PR #327, opened 2022-08-27
Update sponsoring page

veroandreo:
I have started updating sponsoring page. For now, some cleaning, rewording and the addition of students grants as way in which we use the money.

What about adding some quotes from Robert Dzur there too?

On the longer run, a simplified sponsoring/supporting plan as OSGeo has (https://www.osgeo.org/about/how-to-become-a-sponsor/)?

### DIFF
--- a/content/contribute/sponsoring.md
+++ b/content/contribute/sponsoring.md
@@ -6,7 +6,7 @@ layout: "general"
 
 ## Sponsoring the GRASS GIS project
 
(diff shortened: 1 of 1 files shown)


neteler (approved):

neteler:
> What about adding some quotes from Robert Dzur there too?

I'd like the idea but to save time it may go into a different pull request. (?)

veroandreo:
> > What about adding some quotes from Robert Dzur there too?
> 
> I'd like the idea but to save time it may go into a different pull request. (?)

Yes, feel free to merge as is and then I continue in a different PR

PR #332, opened 2022-09-02
donate link in donate dialog

nobeeakon:
## what / why
changed the link in donate dialog to point directly to donate page in opencollective.com instead of the project page  (1 click less to donate)


neteler (approved):
Good idea.

PR #333, opened 2022-09-03
Page title

nobeeakon:
## What
Removed the page title section and the breadcrums

imagen

## Why
Title + breadcrums (that text of `/home/news`) use a lot of vertical space, which makes the content be a bit hidden below it and the information those two elements show may not be adding much.

 - Page title:
    - Most pages have the title within the content. For example: [news](https://grass.osgeo.org/news/), [manuals](https://grass.osgeo.org/learn/manuals/), etc. So it ends up being redundant and takes a lot of vertical space.
 - Breadcrums:
    - There is not much nesting in the site (like` home/learn/manuals/manuals1/manuals2/...`), so may not be add much. And also not many different options at the each url level, for example, `learn/manuals` doesn't group many different subroutes, and same applies for the other pages.
    - Think that primary navigation (menu + links) works just fine.
    - Some breadcrums look more like urls than breadcrums. For example [in news](https://grass.osgeo.org/news/2022_06_05_results_student_grant_caitlin/) the breadcrum says: home/news/2022_06_05_results_student_grant_caitlin , which is not ideal for users
    -


wenzeslaus (approved):
Title: I agree. They don't add much. ...except confusion sometimes since sometimes they don't fit with the first heading.

Breadcrumbs: Given how they work now (URL-like pieces, sometimes not clickable), they take space (although not much).

Diff: Only removes code, that's fine with me.

veroandreo:
For me it's a +0 as I liked the aesthetics of the green title over the light green background... I won't block the merge though

nobeeakon:
> For me it's a +0 as I liked the aesthetics of the green title over the light green background... I won't block the merge though

it's also about the space usage,  for example on mobile you loose around 16% of the view in unnecessary information

neteler:
Just for illustration, some larger screenshots:

Currently on server:
image

After this PR:
image

neteler (approved):

neteler:
Thanks!

PR #338, opened 2022-10-14
User statistics: add matomo javascript code snippet

neteler:
To know more about Web accesses on https://grass.osgeo.org/ and download stats, this PR adds the needed matomo code (https://matomo.org/).

See:
- https://lists.osgeo.org/pipermail/sac/2022-October/014745.html
- https://lists.osgeo.org/pipermail/sac/2022-October/014747.html

@lucadelu kindly created a new matomo for GRASS GIS and me as an initial user. The instance is hosted by GFOSS.it at https://2022.foss4g.org/matomo/

### DIFF
--- a/themes/grass/layouts/partials/footer.html
+++ b/themes/grass/layouts/partials/footer.html
@@ -40,7 +40,7 @@
 <div class="container">
      <div class="row">
       <div class="col-md-5 text-md-left text-center">
-       {{ with .Site.Params.copyright }} <p><small>&copy; {{ replace . "{year}" now.Year }} (<a href="/about/license/">license</a> | <a href="about/privacy">Privacy</a> )</small></p> {{ end }}
+       {{ with .Site.Params.copyright }} <p><small>&copy; {{ replace . "{year}" now.Year }} (<a href="/about/license/">license</a> | <a href="about/privacy">privacy</a> )</small></p> {{ end }}
       </div>
       <div class="col-md-2 text-md-left text-center">
         {{if .IsHome}}
--- a/themes/grass/layouts/partials/footer.html
+++ b/themes/grass/layouts/partials/footer.html
@@ -1,10 +1,48 @@
+<div class="container fixed-bottom ">
+  <div class="m-2">
+      <div id="manage" class="mx-md-2 p-2 bg-white shadow-lg rounded d-none" data-c-view="manage"> <!--  -->
+          <h2>Manage Cookies</h2>
+          <p>We use cookies to provide and secure our website, as well as to analyze its usage, in order to offer the best user experience. You can modify your choice in any moment clicking on "Cookie preferences" button. 
+          </br> To learn more about our use of cookies see our <a href="policy.php">Privacy Policy</a>.</p>
+
+          <div class="my-3">
+              <div class="form-check">
+                  <input data-c-check="necessary" data-c-check-default="true" class="form-check-input" type="checkbox" value="" disabled="" id="necessary">
+                  <label class="form-check-label" for="necessary">
+                      Necessary. Mandatory cookies needed for basic site use
+                  </label>
+              </div>
+
+              <div class="form-check">
+                  <input data-c-check="analysis" data-c-check-default="true" class="form-check-input" type="checkbox" id="analysis">
+                  <label class="form-check-label" for="analysis">
+                      Site traffic analysis. Optional, needed for site access analysis tool
+                  </label>
+              </div>
+          </div>
+
+          <div class="d-flex">
+              <div data-c-action="accept" class="btn btn-warning flex-fill me-1">
+                  Accept
+              </div>
+              <div data-c-action="all" class="btn btn-primary flex-fill">
+                  All
+              </div>
+              <div data-c-action="necessary" class="btn btn-primary flex-fill ms-1">
+                  Necessary
+              </div>
+          </div>
+      </div>
+  </div>
+</div>
+
 <footer class="section bg-gray pb-0">
 <div class="container">
      <div class="row">
-      <div class="col-md-4 text-md-left text-center">
+      <div class="col-md-3 text-md-left text-center">
        {{ with .Site.Params.copyright }} <p><small>&copy; {{ replace . "{year}" now.Year }} (<a href="/about/license/">license</a>)</small></p> {{ end }}
       </div>
-      <div class="col-md-4 text-md-left text-center">
+      <div class="col-md-2 text-md-left text-center">
         {{if .IsHome}}
         <p><small><a href="about/mirrors">Mirror sites</a></small></p>
         {{end}}
@@ -13,8 +51,13 @@
 	<p><small><a href="https://github.com/OSGeo/grass-website/tree/master/content/{{ .Section }}/{{ with .File }}{{ .BaseFileName }}{{ end }}.md" class="fa fa-github"> <span class="edt">Edit this page on GitHub</span></a></small></p>
   {{end}}
       </div>
+      <div class="col-md-3 text-md-right text-center">
+        <div class="widget widget-light">
+          <div id="open" class="btn btn-primary shadow p-1 rounded-pill" data-c-view="button">Cookie preferences</div>
+        </div>
+      </div>
       <div class="col-md-4 text-md-right text-center">	
-	<ul class="list-inline mb-3">
+	      <ul class="list-inline mb-3">
           {{ range .Site.Params.socialIcon }}
           <li class="list-inline-item"><a class="text-color d-inline-block p-2" href="{{ .link }}"><i class="{{ .icon }}"></i></a></li>
           {{ end }}
@@ -28,5 +71,11 @@
 <script>hljs.initHighlightingOnLoad();</script>
 <script src="{{ .Site.BaseURL }}plugins/gallery/photoswipe.js"></script>{{ $script := resources.Get "js/script.js" | minify}}
 <script src="{{ $script.Permalink }}"></script>
+<!-- this is for cookies -->
+<script type="module" src="{{ .Site.BaseURL }}plugins/cookies/cookify_loader.js"></script>
+<script type="text/javascript" data-c-script="analysis">
+    if (window.track) window.track()
+</script>
+
 </body>
 </html>
--- a/themes/grass/layouts/partials/footer.html
+++ b/themes/grass/layouts/partials/footer.html
@@ -39,7 +39,7 @@
 <footer class="section bg-gray pb-0">
 <div class="container">
      <div class="row">
-      <div class="col-md-3 text-md-left text-center">
+      <div class="col-md-4 text-md-left text-center">
        {{ with .Site.Params.copyright }} <p><small>&copy; {{ replace . "{year}" now.Year }} (<a href="/about/license/">license</a>)</small></p> {{ end }}
       </div>
       <div class="col-md-2 text-md-left text-center">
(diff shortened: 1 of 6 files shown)


ninsbl:
Nice!

lucadelu:
@neteler the code for Matomo it self is correct, probably we need to add something for choosing cookies. We used [cookify.js](https://github.com/Jersyfi/Cookify/blob/v3/src/index.js) and some own code (that I can share, if you need)

neteler:
I thought of adding text to the imprint, do you mean that?
But yes, please share :)

neteler:
Probably this would be useful?
https://hugocodex.org/add-ons/cookie-consent/

veroandreo:
Nothing seems broken when building the website from this branch, but what shall I test ? :nerd_face:

lucadelu:
> I thought of adding text to the imprint, do you mean that? But yes, please share :)

I don't think this is enough for the GDPR

lucadelu:
> Probably this would be useful? https://hugocodex.org/add-ons/cookie-consent/

yes this could be a good solution, we need to test if it works properly. probably you need to replace ga.js with matomo.js and copy it in `static/js`

neteler:
> I don't think this is enough for the GDPR

Could this be interesting?
https://matomo.org/blog/2020/02/web-analytics-cookies-gdpr/
--> Cookie-less tracking - disable cookies and ensure cookie compliance

neteler:
Perhaps @nobeeakon could support here? I'm a bit stuck...

nobeeakon:
I think we don't need  super detailed stats so disabling cookies should be fine (only affects [some stats](https://matomo.org/faq/general/faq_156/) ). To be honest I don't know what stats does Matomo provides, so don't know what information we are getting from it but from that link I think it will be fine. Anything I can help @neteler ?

neteler:
Well, we need to have (I believe) this button:

image

Like here: https://2022.foss4g.org/policy.php

Less precise stats are fine. All in all we want to know the download and general visitor stats, that's all.

nobeeakon:
cool @neteler, yep, if no fine grained stats are needed think it is just ok to disable the cookies, thus we don't have to worry about this. Would be nice to see those stats. Also, if needed I can help with the download part.

veroandreo:
> cool @neteler, yep, if no fine grained stats are needed think it is just ok to disable the cookies, thus we don't have to worry about this. Would be nice to see those stats. Also, if needed I can help with the download part.

Indeed! I'd be interested in seeing them, as well :nerd_face:

lucadelu:
@neteler this commit db7295e15f810e0acad51a846e45d37070dbbaeb should add the capabilities to choose cookies, it is rebased on this pull request

neteler:
> @neteler this commit [db7295e15f810e0acad51a846e45d37070dbbaeb](https://github.com/lucadelu/grass-website/commit/db7295e15f810e0acad51a846e45d37070dbbaeb) should add the capabilities to choose cookies, it is rebased on this pull request

Thanks for you contribution!

image

neteler:
The only minor drawback is the layout: the button kicks the (license) to the next line. Maybe too much padding around the button?

image

neteler:
Now it looks better:

image

nobeeakon:
this is real nice, but, do we need it? Like do we need to have the additional information (thinking for example that in the end is not just about  cookies, but also privacy) or is it ok -for our intended use- to have the less grained stats?

neteler:
What about
- disable cookie tracking
- not showing the cookie popup
- keeping the "Cookie preferences" button at bottom, same as on https://2022.foss4g.org/

?

nobeeakon:
think disable cookie tracking should be ok. Don't clearly understand why would the cookie preference button is needed if we are ok with disabling cookies? (also, please note that the [code](https://github.com/OSGeo/grass-website/blob/f7c7bf659e38c0dfb5d098e5c8fcb19354cddfa4/themes/grass/layouts/partials/footer.html#L6) is linking to page we don't  have). My consideration is that if we go with disabling the cookies, that will allow us to not do some additional work like: privacy policy page, additional code maintenance, etc.

petrasovaa:
Just to make sure, the main question is whether we need to track new/unique/returning visitors? The rest of the limitations seems irrelevant for us.

> Matomo uses cookies to store a unique visitor ID, used to recognize visitors from previous visits. When cookies are disabled, Matomo will still be able to determine unique visitors based on IP address and other footprints, but this will be inaccurate. So the following metrics will be inaccurate: Unique visitors, New visitors, Returning visitors.

I agree it may be better to keep it simple for now and reevaluate later if we need more data.

lucadelu:
> this is real nice, but, do we need it? Like do we need to have the additional information (thinking for example that in the end is not just about cookies, but also privacy) or is it ok -for our intended use- to have the less grained stats?

if we want to use matomo, yes it is needed

lucadelu:
> think disable cookie tracking should be ok. Don't clearly understand why the cookie preference button is needed if we are ok with disabling cookies? (also, please note that the [code](https://github.com/OSGeo/grass-website/blob/f7c7bf659e38c0dfb5d098e5c8fcb19354cddfa4/themes/grass/layouts/partials/footer.html#L6) is linking to page we don't have). My consideration is that if we go with disabling the cookies, that will allow us to not do some additional work like: privacy policy page, additional code maintenance, etc.

but if we want to disable cookies we don't need matomo since it use cookies

neteler:
... with this exception?

https://matomo.org/cookie-consent-banners/

neteler:
Now this PR (should, pls check!) disable cookies entirely, see 1d98c2dc838fcb5138de88012a00c41381f0bb96

neteler:
> I think we don't need super detailed stats so disabling cookies should be fine (only affects [some stats](https://matomo.org/faq/general/faq_156/) ). To be honest I don't know what stats does Matomo provides (and more important what information we want to have... and what are we planning to do with it...), so don't know what information we are getting from it but from that link I think it will be fine. Anything I can help @neteler ?

So, no cookies should be enabled. If @nobeeakon (and others) could review the current PR, I'd be happy.

veroandreo:
Most things look good to me, I'd edit `licence` or `Privacy` in the footer so they appear both with uppercase or lowercase:
image

The only page in which the footer height/width seems to be affected is `learn`: 
image

neteler:
@lucadelu how to disable the popup? (the code may remain)

neteler:
> @neteler will take a look later today :)

Thanks! Open problems are:
- no pop-up needed due to use of ` _paq.push(['disableCookies']);` but I don't know how to disable it
- importantly: the matomo code does not appear at all (at least in my tests). Maybe due to e48c0a1b8eb97138145968d1d1fc54a3f62f1a60 in which I intended to remove duplicated code as I thought that `themes/grass/static/plugins/cookies/track.js` does the job?

themes/grass/layouts/partials/footer.html
```diff
@@ -1,10 +1,48 @@
...
+              <div data-c-action="all" class="btn btn-primary flex-fill">
+                  All
+              </div>
+              <div data-c-action="necessary" class="btn btn-primary flex-fill ms-1">
+                  Necessary
+              </div>
+          </div>
+      </div>
+  </div>
+</div>
```
nobeeakon, line 37: if the popup is not needed you can remove this <div> (line 1 - 37)

themes/grass/layouts/partials/footer.html
```diff
@@ -28,5 +71,11 @@
 <script>hljs.initHighlightingOnLoad();</script>
 <script src="{{ .Site.BaseURL }}plugins/gallery/photoswipe.js"></script>{{ $script := resources.Get "js/script.js" | minify}}
 <script src="{{ $script.Permalink }}"></script>
+<!-- this is for cookies -->
+<script type="module" src="{{ .Site.BaseURL }}plugins/cookies/cookify_loader.js"></script>
```
nobeeakon, line 75: no need for this if we are not using cookies

themes/grass/layouts/partials/footer.html
```diff
@@ -13,8 +51,13 @@
 	<p><small><a href="https://github.com/OSGeo/grass-website/tree/master/content/{{ .Section }}/{{ with .File }}{{ .BaseFileName }}{{ end }}.md" class="fa fa-github"> <span class="edt">Edit this page on GitHub</span></a></small></p>
   {{end}}
       </div>
-      <div class="col-md-4 text-md-right text-center">	
-	<ul class="list-inline mb-3">
+      <div class="col-md-2 text-md-right text-center">
+        <div class="widget widget-light">
+          <div id="open" class="btn btn-primary shadow p-1 rounded-pill" data-c-view="button">Cookie preferences</div>
+        </div>
+      </div>
```
nobeeakon, line 58: no need for this button if we are not using cookies (line 54 - 58)

themes/grass/static/plugins/cookies/cookify.js
```diff
@@ -0,0 +1,330 @@
+export default class Cookify {
```
nobeeakon, line 1: this file is not needed if we are not using cookies

themes/grass/static/plugins/cookies/cookify_loader.js
```diff
@@ -0,0 +1,24 @@
+import Cookify from './cookify.js'
```
nobeeakon, line 1: this file is not needed if we are not using cookies

themes/grass/static/plugins/cookies/track.js
```diff
@@ -0,0 +1,20 @@
+export default function track() {
```
nobeeakon, line 1: we could use what they provide: https://developer.matomo.org/guides/tracking-javascript-guide :
replace what is in this file with the stuff they provide

````
  var _paq = window._paq = window._paq || [];
  _paq.push(['trackPageView']);
  _paq.push(['enableLinkTracking']);
  (function() {
    var u="//{$MATOMO_URL}/";
    _paq.push(['setTrackerUrl', u+'matomo.php']);
    _paq.push(['setSiteId', {$IDSITE}]);
    var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
    g.type='text/javascript'; g.async=true; g.src=u+'matomo.js'; s.parentNode.insertBefore(g,s);
  })();
  ```
  
  alternatively, put back this in the header: https://github.com/OSGeo/grass-website/commit/e48c0a1b8eb97138145968d1d1fc54a3f62f1a60

File: themes/grass/layouts/partials/footer.html [hunk 7]
```diff
@@ -28,5 +71,11 @@
 <script>hljs.initHighlightingOnLoad();</script>
 <script src="{{ .Site.BaseURL }}plugins/gallery/photoswipe.js"></script>{{ $script := resources.Get "js/script.js" | minify}}
 <script src="{{ $script.Permalink }}"></script>
+<!-- this is for cookies -->
+<script type="module" src="{{ .Site.BaseURL }}plugins/cookies/cookify_loader.js"></script>
+<script type="text/javascript" data-c-script="analysis">
+    if (window.track) window.track()
+</script>
```
nobeeakon, line 78: guess we could just use what they provide: https://developer.matomo.org/guides/tracking-javascript-guide
alternatively put back this changes https://github.com/OSGeo/grass-website/commit/e48c0a1b8eb97138145968d1d1fc54a3f62f1a60

neteler:
@nobeeakon what if we close this PR and start afresh with a new and clean one? I'd happily to leave it to you (given my JS foo) :-)

nobeeakon:
@neteler sounds good to me :)

nobeeakon:
done in #342

PR #349, opened 2023-01-29
Add .hugo_build.lock and hugo.exe file to .gitignore

echoix:
To help out newcomers when running `hugo server` locally, this prevents showing that this is a new file that would be added to the repository.

It's not much, but it was a common commit I was doing when exploring how to make contributing easier in the repo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,6 @@ resources/
 
 # Temporary lock file while building
 /.hugo_build.lock
+
+# Ignore the Windows Hugo executable
+hugo.exe
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 .DS_Store
 public/
 resources/
+
+# Temporary lock file while building
+/.hugo_build.lock


.gitignore
```diff
@@ -1,3 +1,6 @@
 .DS_Store
 public/
 resources/
+
+# Temporary lock file while building
+/.hugo_build.lock
```
neteler, line 6: Are you sure that `/` is right here? Elsewhere I see it without ...

echoix: Take a look at 

https://github.com/github/gitignore/blob/4488915eec0b3a45b5c63ead28f286819c0917de/community/Golang/Hugo.gitignore
Or 
https://github.com/toptal/gitignore/blob/2d56fdde1e84eaeb6688cc3536afdfd3c54c6016/templates/Hugo.gitignore


In the gitignore documentation, the first bullet in the section examples linked below, explain that it is to match only at the root, not inside subdirectories. 
https://git-scm.com/docs/gitignore/2.34.0#_examples

I could retry and make a test that a file called `.hugo_build.lock` stored in the static folder (if by some random chance we would want to make a download named like that), to make sure it isn't ignored.

echoix:
When I was helping out @withgaurav offline this morning, I understood why other templates ignore hugo.exe in the root of the repo. It is quite easy to tell users to download a specific (or latest) hugo version and copy the single file hugo.exe from the compressed folder into the root of the repo, and only run `./hugo.exe server` to collaborate on windows.

Should we update the gitignore proactively before updating some contribution help?

.gitignore
neteler: Ok, I see. Fine for me.

veroandreo:
> When I was helping out @withgaurav offline this morning, I understood why other templates ignore hugo.exe in the root of the repo. It is quite easy to tell users to download a specific (or latest) hugo version and copy the single file hugo.exe from the compressed folder into the root of the repo, and only run `./hugo.exe server` to collaborate on windows.
> 
> Should we update the gitignore proactively before updating some contribution help?

Do you mean also adding hugo.exe to .gitignore?

echoix:
Yep, I didn't adjust the PR yet, I'm only on a cellphone for now.

echoix:
Ok, if there isn't anything else blocking, I think it's pretty much finished :)

PR #350, opened 2023-01-29
Fixed broken links for manual pages

withgaurav:
I was opening manual pages, but it was giving me error. So I have tried to fix it

### DIFF
--- a/content/learn/_index.en.md
+++ b/content/learn/_index.en.md
@@ -11,7 +11,7 @@ weight: 2
 <h3 class="mt-2 mb-4">Plenty of resources</h3>
 
 There are plenty of resources online to get started and learn GRASS GIS: from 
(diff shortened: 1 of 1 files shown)


neteler:
Sorry to say, but this is not correct. There must be no absolute link here, otherwise the mirror sites will not work.

What was exactly the error?

withgaurav:
image


This was the error.

veroandreo:
That's because `manuals` is not a folder but a .md file within `content/learn` that points to the manual pages in the osgeo server. Building the site locally and clicking over `learn/manuals` works fine. What were you trying to do when facing this error? I am not an hugo expert, but I can tell that studying the structure of the website to understand how things are related, building and testing minor changes locally is really helpful (and I strongly recommend it ;-)).

Closing as invalid.

PR #352, opened 2023-02-01
updated colors of learn page

withgaurav:
- I have seen that before hovering text was light in color & after hovering text was dark in color. 

image

- So I have made some changes now the text will look dark in color in color before hovering and light in color after hovering
image

### DIFF
--- a/themes/grass/assets/css/style.css
+++ b/themes/grass/assets/css/style.css
@@ -361,7 +361,7 @@ table img {
   display: block;
   line-height: 1.2em;
   font-size: 2.6em;
-  color: #dde9cc;
+  color: var(--grass-color-dark);
 }
 .htl:hover {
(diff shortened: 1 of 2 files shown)


veroandreo:
I have tested now, but the behavior does not match what you have written above. The color does not change with hovering, it remains the same, i.e. `var(--grass-color-dark)`. Is that what you intended or meant?

Perhaps we should homogenize the behavior of this page with links all over the site, that they appear as `--grass-color` and change to `--grass-color-dark` when hovering. What do you think?

withgaurav:
@veroandreo Great, it sounds like a good idea to homogenize the behavior of links on the site, so that they all appear as --grass-color and change to --grass-color-dark when hovering. This will make the site look more consistent and visually appealing.

PR #353, opened 2023-02-05
404.html: advertise the current GRASS GIS version

### LINKED ISSUE #351: How to use the grass version number variables in HTML?
(this related to "fix/grass version numbers #318")

While I know how to use the variables in .md files, how to use it in .html?

Specifically, I want to add a line with the current version number to `themes/grass/layouts/404.html` since I see from our Matomo statistics that ppl try to find very old versions (they go to old news for whatever reason, then fail to find very old binaries) and need a hint about the current version on the "404" page:

```
            <br><br>
            Current version: GRASS GIS {{< grassVersion version="current" >}}
```

Now, `hugo server` dislikes this addition (as it is likely in markdown-only style).

Any pointer on usage of the variables in HTML files would be welcome.

neteler:
Aim: reduce confusion when visitors land on an outdated version announcement (problem identified with Matomo, search behaviour analysis). At least, now they know what to look for....

Shortcode from: https://github.com/OSGeo/grass-website/tree/master/themes/grass/layouts/shortcodes

## old
image

## new
image

Fixes #351

### DIFF
--- a/themes/grass/layouts/404.html
+++ b/themes/grass/layouts/404.html
@@ -7,11 +7,14 @@
       <div class="row">
         <div class="col-lg-8 text-center mx-auto">
           <h1><a href="{{ "/" | relURL }}"><img alt="GRASS GIS" src="{{.Site.BaseURL}}/images/logos/grasslogo.svg" width="33%"></a></h1>
-	  <div class="alert alert-warning"><h2 style="margin:10px;color:red;">Sorry, there's nothing here.<span style="display:block;margin:10px;font-size:.75em;">
+          <div class="alert alert-warning">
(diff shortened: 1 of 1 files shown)


veroandreo (approved):

echoix:
I'm still concerned here as why there are some outdated version announcements that creates 404. What articles are concerned that get removed when a version is released?

neteler:
We do not remove any article but some very old Windows installers have been removed due to disk space limitations.

echoix:
Oh, ok, so the 404s are for download links only.

neteler:
Right.
Plus used for catching outdated links on external pages which (we cannot modify) pointing to non-existing pages on grass.osgeo.org.

PR #371, opened 2023-06-06
Docker download: Update docker organization

mmacata:
As the GRASS GIS docker images are currently moving from "mundialis" to [OSGeo dockerhub organization](https://hub.docker.com/u/osgeo), this PR updates it.


mmacata:
(First images are on Dockerhub and image sizes updated)

On the website, the docker image tag is referenced with a specific version, e.g.
```
docker pull osgeo/grass-gis:8.2.1-alpine
```
On Dockerhub this image is not yet available but published as `docker pull osgeo/grass-gis:stable-alpine`.
It will only appear when there is a release. So my question is, maybe to @neteler, when is this website build? Should this PR stay in a draft stage until there is a g82 release?

veroandreo:
Hi Carmen! Thanks for the changes!

I'd suggest to remove the version number from the website if that does not exist on Dockerhub, to avoid confusion when people copy-paste from the website. 

In any case, I find it difficult to match the terms used in the website for the download section (current, legacy, preview) with the [tags](https://hub.docker.com/r/osgeo/grass-gis/tags) in Dockerhub (stable, latest, releasebranch83). How do they relate? Which one to keep? IIUC, releasebranch83 will become stable once 8.3 is released and latest will still be main, right?

mmacata:
Hi Vero, thanks for the feedback :)

> Hi Carmen! Thanks for the changes!
> 
> I'd suggest to remove the version number from the website if that does not exist on Dockerhub, to avoid confusion when people copy-paste from the website.

The versions will appear on Dockerhub when a release is made. Just I don't know the correlation between a release and the website, e.g. when the variable `grass.current_version` will be updated in this repository. 
> 
> In any case, I find it difficult to match the terms used in the website for the download section (current, legacy, preview) with the [tags](https://hub.docker.com/r/osgeo/grass-gis/tags) in Dockerhub (stable, latest, releasebranch83). How do they relate? Which one to keep? IIUC, releasebranch83 will become stable once 8.3 is released and latest will still be main, right?

Yes, that was the idea. `latest` always points to main and `stable` looks at the latest release. There was just a tiny discussion about the docker image tags but in the end they just stayed the same as before at `mundialis/grass-py3-pdal`. I slightly had the tendency to drop the `stable-XXX` and `latest-XXX` tags and simply name them after the branches, it would be `main-XXX` and `releasebranch_8_2-XXX` then and it would be clear what's inside. But according to https://hub.docker.com/repository/docker/mundialis/grass-py3-pdal/tags?page=1&ordering=last_updated they are used frequently and users don't need to update the tag names.

What is your opinion on the tags?

wenzeslaus:
I think the expected behavior in Docker world is that there is latest which points to whatever is the default stable version.

echoix:
First of all, it's possible to have multiple tags for each image, to accommodate different granularity or anything else. Next, it might be more useful to have also the version as semver at the beginning of the tag, and have after a dash a suffix with the remaining info. That would enable users to have their dependencies updated by tools, like renovate that we use. Take a look at the docs where they explain what they support (that is the common usage of tags): https://docs.renovatebot.com/docker/#docker 

Again, I'm only talking about adding a tag, not replacing what is done at the moment, only to accommodate more use cases and match more expected behavior.

mmacata:
Thanks for your opinions. I will leave this open until a release (not pre-release as was done recently) will be made, then also tags with a version will appear on dockerhub and might add some ideas to the discussion.

neteler:
I have edited the PR and tried to add "wxgui" entries.

Unfortunately I have no idea how to get the new shortcode `currentVersionUnderscore` working, seems that  70ee76f  is incomplete. Does @nobeeakon have a suggestion here?

image

veroandreo:
> I have edited the PR and tried to add "wxgui" entries.
> 
> Unfortunately I have no idea how to get the new shortcode `currentVersionUnderscore` working, seems that [70ee76f](https://github.com/OSGeo/grass-website/commit/70ee76f0c38beadecdec2b654c7c71cc4bac0d97) is incomplete. Does @nobeeakon have a suggestion here?
> 
> image

Not sure how, but I guess something needs to be added here as well: https://github.com/OSGeo/grass-website/blob/master/themes/grass/layouts/shortcodes/grassVersion.html

veroandreo:
> I think the expected behavior in Docker world is that there is latest which points to whatever is the default stable version.

IIUC, this is then not the same than latest tag pointing to main branch which is a dev branch, not stable. I do like latest and stable as tags, but let's agree on their meaning and perhaps also think again about the terminology we use within the website download section.

neteler:
The new releases shows up as expected:

https://hub.docker.com/r/osgeo/grass-gis/tags

neteler:
> On Dockerhub this image is not yet available but published as `docker pull osgeo/grass-gis:stable-alpine`.
> It will only appear when there is a release.

@mmacata As of today (so, after the G8.3.0 release) it still shows 8.2, not sure why?

```
docker run -it --rm osgeo/grass-gis:stable-alpine grass --version
GRASS GIS 8.2.2dev
...
```

echoix:
> > On Dockerhub this image is not yet available but published as `docker pull osgeo/grass-gis:stable-alpine`.
> 
> > It will only appear when there is a release.
> 
> 
> 
> @mmacata As of today (so, after the G8.3.0 release) it still shows 8.2, not sure why?
> 
> 
> 
> ```
> 
> docker run -it --rm osgeo/grass-gis:stable-alpine grass --version
> 
> GRASS GIS 8.2.2dev
> 
> ...
> 
> ```

If you're running on a system where you already have the image tag locally, it won't pull again when starting it with run. You can use the following to pull the image again
```

docker pull osgeo/grass-gis:stable-alpine

```

mmacata:
There is no mapping of `stable` in the github workflow which pushes the Dockerimage:
https://github.com/OSGeo/grass/blob/main/.github/workflows/docker.yml#L67
So this would need to be updated.

neteler:
Indeed:

```
find . -type f | xargs grep releasebranch_8_2
...
./.github/workflows/docker.yml:# * creates tags stable-alpine, stable-debian and stable-ubuntu for releasebranch_8_2
./.github/workflows/docker.yml:  # For releasebranch_8_2, created tags are:
./.github/workflows/docker.yml:          elif [ "$GITHUB_REF" == "refs/heads/releasebranch_8_2" ]
./docker/alpine/README.md:git checkout remotes/origin/releasebranch_8_2
./docker/debian/README.md:git checkout remotes/origin/releasebranch_8_2
./docker/ubuntu/README.md:git checkout remotes/origin/releasebranch_8_2
./docker/ubuntu/README.md:git checkout remotes/origin/releasebranch_8_2
./docker/ubuntu_wxgui/README.md:git checkout remotes/origin/releasebranch_8_2
./docker/ubuntu_wxgui/README.md:git checkout remotes/origin/releasebranch_8_2
./singularity/debian/README_debian.md:git checkout remotes/origin/releasebranch_8_2
```

mmacata:
Suggestions for a slighly different tag approach in this PR: https://github.com/OSGeo/grass/pull/3075
Discussion welcome.

neteler:
Since https://grass.osgeo.org/download/docker/ is pointing to meanwhile non-existing docker repos, any chance to get this merged?

mmacata:
So the related PR was updated for `latest` tag to point to the last release, `stable-*` was removed and `current-*` introduced. So it is more in line with the GRASS GIS terms.
Also the `*-wxgui` errors were fixed.

neteler:
Great work! Merging.

wenzeslaus:
The plan is to keep the branches or replace them with tags for 8.3? The section headings use x.y.z, but then the info box and commands are for a branch. While using a release branch is perfectly fine from GRASS project perspective, people less familiar with the GRASS development life cycle will likely prefer to use a release because that seems more stable. 7.8 has release in heading and commands, but the info box mentions a branch.

PR #381, opened 2023-08-18
news: mastodon news feed support added

### LINKED ISSUE #380: Twitter feed not working
The twitter feed is no longer working. Check: https://grass.osgeo.org/news/

image

neteler:
This PR ~~replaces~~ adds on top of the broken X feed our mastodon news feed from https://fosstodon.org/@grassgis .

Code kindly crafted with https://www.mastofeed.com/

Screenshot:

image

May replace X (formerly Twitter) since their feed is broken if not login'ed to X.
Fixes #380

### DIFF
--- a/themes/grass/layouts/events/allevents.html
+++ b/themes/grass/layouts/events/allevents.html
@@ -39,9 +39,13 @@
       </div>
       </div>
       <div class="col-md-4 bg-white">
-        <h3 class="mt-20"><a href="#">@GRASSGIS</a></h3>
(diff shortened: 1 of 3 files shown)


veroandreo:
Thanks @neteler !! Here it works in Chrome when I'm logged in to X, but not in Firefox (whether I'm logged in or not)... I have no clue. Anyway, I like the idea of having both feeds. Maybe once we grow the audience in Mastodon, we can drop X... So, +1 for merging :)

PR #389, opened 2023-09-09
Update mac.en.md

cmbarton:
Fixed download links to GRASS for Mac web page

### DIFF
--- a/content/download/mac.en.md
+++ b/content/download/mac.en.md
@@ -33,7 +33,7 @@ or install the latest available version from <a href="https://ports.macports.org
 <ul>
 <li>
  {{< donateDialog isToggle=true >}}  
-    <a href="https://cmbarton.github.io/grass-mac/_pages/download-grass/"><i class="fa fa-download"></i> Download </a>
+    <a href="[https://cmbarton.github.io/grass-mac/download/](https://cmbarton.github.io/grass-mac/download/)"><i class="fa fa-download"></i> Download </a>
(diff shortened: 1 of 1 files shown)


content/download/mac.en.md
```diff
@@ -33,7 +33,7 @@ or install the latest available version from <a href="https://ports.macports.org
 <ul>
 <li>
  {{< donateDialog isToggle=true >}}  
-    <a href="https://cmbarton.github.io/grass-mac/_pages/download-grass/"><i class="fa fa-download"></i> Download </a>
+    <a href="[https://cmbarton.github.io/grass-mac/download/](https://cmbarton.github.io/grass-mac/download/)"><i class="fa fa-download"></i> Download </a>
```
veroandreo, line 36:
```suggestion
    <a href="https://cmbarton.github.io/grass-mac/download/"><i class="fa fa-download"></i> Download </a>
```
The fix was mixing html and md links formatting

cmbarton: Good point Vero. I was being careful to paste in the link in the format currently used. Do you think that markdown would be better or more reliable here?  If so, we might want to change the entire line to markdown if we can keep the buttons.

veroandreo: Nope, I'd leave it in html that allows us to use fontawesome icons easily

cmbarton:
Hopefully this will fix the links.

Issue #402, opened 2023-11-27
Update fontawesome version 


manasrai1406:
As mentioned in the comment on issue : #392 the fontawesome library should be upgraded to support new twitter, facebook, youtube icons.
I have worked on this issue and have confirmed that with v6 the new icons are displaying correctly as can be seen in the screenshot below

![image](https://github.com/OSGeo/grass-website/assets/106554051/eb1e9710-8479-46d0-b2d2-85a1b658c614)

As this change requires updating the library which will result in massive file differences, I am creating this issue to seek permission to do so.

I am ready to open a PR as soon as I get the green light.

PR #403, opened 2023-11-27
Update fontawesome to v6

manasrai1406:
As mentioned in the said issue, I've updated the fontawesome version to v6.
Instead of implementing the whole fontawesome kit, I've stripped down the folder massively from 2k+ changes to only 51.
If any files need to be removed or restored, please do let me know.
  
fixes: #402


veroandreo:
Thanks! :clap: 

I tested locally and everything seems to compile fine, only those icons mentioned by @petrasovaa need to be fixed

petrasovaa (approved):

Issue #424, opened 2024-03-26
Content header cutoff on mobile


cwhite911:
Content headers are getting cut off on mobile devices. I've not looked to see which layouts are impacted yet.

![image](https://github.com/OSGeo/grass-website/assets/3803800/0ca42c00-f2c3-4d62-95d5-ab399396b6b7)

echoix:
It's funny, I made a screenshot on Friday for something similar, on Safari with an iPhone on iOS 17.4. 

![image](https://github.com/OSGeo/grass-website/assets/27212526/8fae9beb-539e-4414-afda-48abe220d840)

Its possible to scroll up to see the full title, until you lift your finger and goes back to this position.

cwhite911:
Fixed the root cause of this in PR https://github.com/OSGeo/grass-website/pull/425

PR #425, opened 2024-03-26
issue-424: Fixing iscutoff content issue

cwhite911:
I've starting to resolve issue #424 by updating the mt-95 and section classes.

Additional, issues with some pages html syntax were also updated which were causing undesired interactions between the banner, header, and page content.


veroandreo (approved):
Looks good!

PR #452, opened 2024-08-09
Gallery: move outdated GUI screenshots to archive subdir

neteler:
https://grass.osgeo.org/learn/gallery/ currently shows (mostly) very old screenshots.

This PR moves outdated files into `archive/` subdirectory to keep them for historical reasons. The remaining GUI screenshots are shown (still their order is rather unclear as new ones should be on top).

New screenshots should be added in a separate PR, see #453


veroandreo:
Thanks @neteler !

PR #468, opened 2024-08-30
Replace 'founding member' by 'project of OSGeo'

wenzeslaus:
Use project of OSGeo rather than founding member becuase the terminology is OSGeo project and that's also what needs to be highlighted.

The founding member has historical significance so it is kept in the history page, but removed from anywhere else.

### DIFF
--- a/content/contribute/sponsoring.md
+++ b/content/contribute/sponsoring.md
@@ -11,7 +11,7 @@ layout: "general"
 ### Why sponsor?
 
 *The GRASS GIS project is a volunteer-based organization*, 
-it is an affiliated project and a project of the
+it is a project of the
 Open Source Geospatial Foundation ([OSGeo](https://osgeo.org/)).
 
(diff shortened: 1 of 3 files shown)


content/contribute/sponsoring.md
```diff
@@ -11,7 +11,7 @@ layout: "general"
 ### Why sponsor?
 
 *The GRASS GIS project is a volunteer-based organization*, 
-it is an affiliated project and founding member of the
+it is an affiliated project and a project of the
```
veroandreo, line 14:
```suggestion
it is a project of the
```
just to simplify...

veroandreo (approved):

content/contribute/sponsoring.md
wenzeslaus: That was definitively the intention just like in the other place.

veroandreo (approved):

PR #476, opened 2024-10-01
Update after moving grass-dev to Discourse

petrasovaa:
image


wenzeslaus (approved):
This looks good as long as this is not the permanent solution and we rethink this after all migrations are done.

PR #481, opened 2024-11-23
Fix broken link and add notes

itsadityaaaaa:


itsadityaaaaa:
Is it the right way to fix broken links??

neteler (approved):
Thanks for your contribution, looks good to me (tested locally).

veroandreo (approved):
Thanks for your contribution, @itsadityaaaaa. Yes, the fix looks good! I had trouble finding a link to the page you fixed from the homepage.

In the future, please create a new branch in your local repository instead of using the main branch. You can refer to this guide in the main GRASS repository that we use for all other grass-related repositories: https://github.com/OSGeo/grass/blob/main/doc/development/github_guide.md.

Thanks again!

PR #508, opened 2025-03-18
Update after migration to discourse forum

petrasovaa:
* Modified community page to add discourse forum
* Put mailing lists towards the end of the page
* Removed GIS on the community page as part of the edits
* Add Discourse to footer

### DIFF
--- a/themes/grass/layouts/partials/banner.html
+++ b/themes/grass/layouts/partials/banner.html
@@ -1,5 +1,0 @@
-<div class="gitter-banner"
-          >
-    Feeling chatty? <a href="https://gitter.im/grassgis/community" target="_blank"
-    >Join our Gitter chatroom</a> if you have any question or want to chat about the software!
-</div>
--- a/themes/grass/layouts/partials/banner.html
+++ b/themes/grass/layouts/partials/banner.html
@@ -1,0 +1,5 @@
+<div class="gitter-banner"
+          >
+    Feeling chatty? <a href="https://gitter.im/grassgis/community" target="_blank"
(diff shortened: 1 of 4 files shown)


wenzeslaus (commented):
This looks great. I'm only worried about all the pages including partial if it does not exist. I don't see a mention of this in the Hugo doc. Maybe an empty file or an empty div?

themes/grass/layouts/partials/banner.html
```diff
@@ -1,5 +0,0 @@
-<div class="gitter-banner"
```
nobeeakon, line 1: would be good to also remove the css class in case is not used anymore

petrasovaa: I was not sure if we want to remove the banner completely, maybe we would want to highlight something else in the future? So I went with minimal changes...

nobeeakon: should be a quick fix (just some css declarations [here](https://github.com/OSGeo/grass-website/blob/master/themes/grass/assets/css/style.css#L676)). Issue being that css tends to accumulate dead code that at some point would need to be removed :)

wenzeslaus:
To remove or not to remove: I would almost be for removing it completely - we can always find it the history, but it would be better to have it readily available when we want to do another banner with different content, e.g. a funding camping. Enabling the banner after complete removal is much more involved since at this point, it needs to go to different templates.

While I agree with the original idea of including it, I suggest to remove the change from this PR to get the other clear changes in.

wenzeslaus (approved):

Issue #513, opened 2025-05-08
[Bug] SEO issues on website


neteler:
We are currently facing the problem that the GRASS website has dropped significantly in the Google search rankings (and probably in other search engines as well).

I have run a SEO (Search Engine Optimization) analysis today with the outcome below.

The tools used are
- https://seorch.de/
- https://search.google.com/search-console?resource_id=https%3A%2F%2Fgrass.osgeo.org%2F&hl=en

These issues should be addressed as soon as possible.

## Meta Description

The meta description is too long: 218 characters.

Target: It should have a maximum of 160 characters. 

## Text - <body> tag

- The website has too little text: 212 words.
- The text length should exceed 400 words. 
- About 10% of the text consists of filler words. perfect
- The text has less than 4 complete sentences. you can do better
- The average sentence length is 21 words. This is too much.

## Hierarchy of titles

```
h1 (No Text)   <-- bug
  h2 (missing)  <-- bug
     h3 download
     h3 learn
     h3 support
     h3 contribute
  h2 latest news
     h3 grass refreshed branding
     h3 grass developer summit 2025
     h3 grass 8.4.1 released
```

Incomplete hierarchy is commonly penalized by search engines.

## Semantic analysis: Readability of the website

(calculated according to Flesch Reading Ease)

- Flesch value: 44
- Readability: difficult
- This is roughly equivalent to a legal text.

## Code / Text Rate

- Text rate: 7.51 %
- The text rate should be above 25 %.

We simply need more text.

## HTML error in source code

HTML not valid. 16 errors found (tested with [W3C Validator](https://validator.w3.org/nu/?doc=https%3A%2F%2Fgrass.osgeo.org%2F))

## Images - <img> Tags

- 3 <img> Tags ohne ALT Attribut - problem
- 10 <img> tags without TITLE attribute - problem

## Open Graph Protocol

- 4 Open Graph attributes found.
- Problem: some required attributes were not found: `og:type missing`

## Largest Contentful Paint (LPC): Pre-load the largest paint image

Info: https://web.dev/articles/lcp

If the LCP element is dynamically added to the page, you should preload the image to improve the LCP.
Affected files and functions:

    https://grass.osgeo.org/images/logos/banner.jpg

banner.jpg: `JPEG image data: 1899x1268 pixels`

[...]

PR #514, opened 2025-05-08
Add canonical URL

petrasovaa:
@cwhite911 could check if this makes sense or it needs to be different? Related to #513 

Scope: avoid that mirror sites take a higher search engine ranking by pointing users to the main site.


wenzeslaus (approved):

wenzeslaus (changes requested):
Wait, this had Permanent link last time I saw this. This will just go to the canonical main page, not the canonical page for each page.

neteler:
> This will just go to the canonical main page, not the canonical page for each page.

Yes, confirmed (I checked this PR out and tested it locally).

neteler:
I think I got it, replacing `.Site.BaseURL` with `.Permalink` in 09570a2dd46b2ae77ee8fdcf8e59920ba2ff87ec.

"Sponsoring" page example:

image

PR #538, opened 2025-05-20
Broken example URLs 

mabdazzam:
fixed broken URLs to example vector and raster modules in C on the manuals page.


veroandreo (approved):
Thanks @mabdazzam !

Issue #547, opened 2025-05-29
design: Fix default link colors


cwhite911:
The link colors are current set to use the secondary color, but the should use the primary color.

echoix:
Thanks for filing it in, we noticed it during the summit, and indeed links in the text are very hard to discern

cwhite911:
Thanks! I have a bunch of updates I'm going to be making now to finish synchronizing the website with the the Docs and tutorials. I was saving the website for last because it is the most complicated.

With that said if you see things that look wrong please add it as a sub issue to the parent of this issue to make sure it's on my radar.

PR #549, opened 2025-06-20
style: changed link colors to primary

cwhite911:
Change the link colors to primary to match the new documentation (Issue #547).

### DIFF
--- a/themes/grass/assets/css/style.css
+++ b/themes/grass/assets/css/style.css
@@ -32,10 +32,10 @@
   --gs-primary-light-color--lightest: rgba(138, 233, 145, 0.071);
 
   /* Primary Dark Color */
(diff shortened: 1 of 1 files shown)


veroandreo:
Thanks @cwhite911! This harmonization will take some changes after this one is merged, it seems. I dunno if it is just the style or the inconsistent use of links and section title levels because of the former style... This one looks too green now :face_in_clouds: Perhaps we don't need to use link in the text if we already have the button, no? 
 
image

Anyway, thanks again!

cwhite911:
@veroandreo  that is a part of the challenge of the greater synchronization of the website styling. We have inconsistent usage of css classes, classes using !important, html elements, and html elements with that use the style attribute. My goal is to rewrite our `style.css` to make it easier to be more consistent.

PR #551, opened 2025-07-03
brand: Added anchor links to brand logo drop down buttons

cwhite911:
I added anchor links to the dropdown buttons on the brand page to prevent the page from navigating to an undefined page if a user second clicks on the button.


petrasovaa:
I don't know if it's my version of hugo or something, but I get this:
image

PR #552, opened 2025-07-03
devcontainer support

cwhite911:
~~I've add Docker support for running the Hugo server, integrating SCSS for theming, and updating dependencies. Below is a breakdown of the most important changes:~~

~~### Development Environment Updates:~~
~~* Added `HUGO_VERSION` to the `.env` file to specify the Hugo version used in Docker (`0.139.4`).~~
~~* Updated Node.js version in `.nvmrc` from `22.14.0` to `v23.3.0`.~~
~~* Added instructions in `README.md` for running Hugo locally with Docker, including commands for installing Node dependencies and launching the Hugo server.~~

~~### Docker Integration:~~
~~* Introduced a `docker-compose.yml` file to define a service for running the Hugo server with the `hugomods/hugo:dart-sass` container, including volume mappings and port configuration.~~

~~### Styling and Theming Enhancements:~~
~~* Added a new SCSS file (`themes/grass/assets/sass/main.scss`)~~
~~* Updated `themes/grass/layouts/partials/head.html` to use the SCSS file (`main.scss`) with Dart Sass for CSS preprocessing, replacing the previous static CSS file.~~

~~### Dependency Management:~~
~~* Added Bootstrap (`^5.3.7`) as a dependency in `package.json` to support modern styling and layout capabilities.~~


echoix:
Sounds like something that would be appropriate to use devcontainers (see the specifications). It can use a docker image directly too, so you’re not that way off

also, using node 23 is kinda weird, node 24 is going to be the LTS at any point now. Otherwise node 22 is used

echoix (commented):
Are the scss and html changes related to this PR?

echoix:
I don't know where the nginx is mentioned, but looks fine.

The version inconsistencies I was mentioning were between the devcontainer file and the other files that you use, that would be the current versions as you said.

Note that I think the website is still built on the Debian server if I'm up to date (are we pushing the artifacts there now?), so we cannot update all the versions we want without planning.


At least for devcontainer, renovate will help keeping them updated. https://docs.renovatebot.com/modules/manager/devcontainer/

The `.nvmrc` file too: https://docs.renovatebot.com/modules/manager/nvm/

Docker compose: https://docs.renovatebot.com/modules/manager/docker-compose/

cwhite911:
> I don't know where the nginx is mentioned, but looks fine. The version inconsistencies I was mentioning were between the devcontainer file and the other files that you use, that would be the current versions as you said.

Wasn't thinking. I changed them locally but they are not committed.

> At least for devcontainer, renovate will help keeping them updated. https://docs.renovatebot.com/modules/manager/devcontainer/
The .nvmrc file too: https://docs.renovatebot.com/modules/manager/nvm/
Docker compose: https://docs.renovatebot.com/modules/manager/docker-compose/

Cool, I'll add these.

> Note that I think the website is still built on the Debian server if I'm up to date (are we pushing the artifacts there now?), so we cannot update all the versions we want without planning.

Yes. I believe that is right. That is why I want to setup docker-compose replicate that environment the best we can for testing the built artifacts.

neteler:
> Note that I think the website is still built on the Debian server ...

Correct

> (are we pushing the artifacts there now?),

You mean tarballs and such?

> so we cannot update all the versions we want without planning.

What do you need to be updated on the server - can I help?

echoix (commented):
Why is the devcontainer PR still containing so much website/css/javascript/content changes?

echoix:
npm ci means clean install. It flushes existing node_modules and only uses the package-lock.json. If you didn’t have the package-lock up to date, that’s why it might not have worked. It would usually be the recommended way, npm install would change to install different dependencies over time. If the setup doesn’t work with npm ci, it will become a problem sooner or later

cwhite911:
I've decided to to merge this PR with PR #599. Initially,  I though they would be better independent but because the PR is also updating the dependencies that require css changes I want to include the scss now to avoid double work of updating the css and then converting it to scss later.

cwhite911:
Closing as superseded by #599. The devcontainer and Docker-support commits from `docker-support` were merged into the `hugo-dart-scss` branch, so #599 already includes this work and carries it forward. Thanks!

Issue #581, opened 2025-12-19
Brand page: download buttons don't work


petrasovaa:
The buttons on the Brand page don't do anything, no drop down, so you can't download anything.

<img width="566" height="114" alt="Image" src="https://github.com/user-attachments/assets/9883be57-45bc-4228-b630-7d2ff4c86ebb" />

cwhite911:
I'm thinking this related to PR #560 where we update Bootstrap 5 without addressing any breaking changes.

kartik-singhhh03:
Hi! @petrasovaa ,
 I’d like to work on this issue.
I believe this is related to Bootstrap 5 dropdown attribute changes (data-toggle → data-bs-toggle).
I’ll open a PR to update the brand page download dropdowns accordingly.

PR #583, opened 2025-12-24
 #581 Update brand download dropdowns to Bootstrap 5 namespaced attributes

subhu339:
#581 
Updated the brand download dropdown buttons to use Bootstrap 5 syntax. The buttons were previously using data-toggle, which is no longer supported; changing these to data-bs-toggle restores the dropdown functionality on the Brand page.

Modified brandDownloadIconDropdownButton.html to use namespaced Bootstrap 5 attributes.


cwhite911:
We will test this as part of the migration Issue #600

cwhite911:
This was addressed during the larger conversion of the site to Bootstrap 5

PR #597, opened 2026-02-05
release history: add new GRASS 85 branch creation tag

neteler:
New GRASS 85 branch creation tag added


cwhite911:
The website currently has two bootstrap versions running.  Currently, the older version of Boostrap (v3) (uses jquery) is loaded first and then a newer version of bootstrap (v5) (doesn't use query) is loaded on top. The collapse function was from Bootstrap v3. I recommend focusing on getting PR #552 merged soon, so we have clearly defined project dependencies allowing us to properly fix these UI bugs and mixed dependencies.

Issue #600, opened 2026-02-06
Migrating to Bootstrap v5 from v3


cwhite911:
We have ad-hoc started the process of migrating to Bootstrap v5 from Bootstrap v3. Overall, I think updating to Bootstrap 5 is good, however, this mixing of dependencies has caused quite a few breaking changes on the site placing us in a bit of a triage mode.

Related open Issues and PRs:
- #583
- #581
- #580 
- #579 
- #575

My suggestion to resolve is to quickly get PRs 

- #552 
- #599 

merged to normalize dependencies and finalize the switch to Bootstrap 5. We can then migrate old Bootstrap 3 and sometime 4 references safely.

Here are the migration guidelines from:

* [Bootstrap 3 to 4](https://getbootstrap.com/docs/4.0/migration/)
* [Bootstrap 4 to 5](https://getbootstrap.com/docs/5.0/migration/) 

Here are the big breaking changes 

* Dropped jQuery.
* Upgraded from Popper v1.x to Popper v2.x.
* Replaced Libsass with Dart Sass as our Sass compiler given Libsass was deprecated.

## Suggested Issues

1. **Inventory current Bootstrap usage and version conflicts**

- Audit templates and partials to identify Bootstrap 3/4 classes, data attributes, and JS usage.
- Key Bootstrap 3/4 patterns currently in the theme:
  - jQuery + Bootstrap bundle loaded together in head: head.html (jquery.slim.min.js, bootstrap.bundle.min.js concatenated), implying reliance on jQuery-based data-API.
  - Navbar uses BS3/4 attributes and classes: navigation.html (data-toggle, data-target, .navbar-toggler, .navbar-collapse, .navbar-nav, .dropdown-menu), and uses ml-auto (BS4 spacing).
  - Tabs use BS4 data-API: tabs.html (data-toggle="tab", .nav-tabs, .tab-pane, .fade, .show).
  - Collapse/Accordion usage with BS3 data-API: script.js ($("#news .panel-collapse:first").collapse();) and gallery toggler uses jQuery events.
  - Grid classes with BS3/4 naming: multiple layouts (e.g., allnews.html, index.html, team.html, list.html) use .col-lg-*, .col-sm-*, .row, .container; spacing utilities like .mt-5, .mb-4 (BS4 utility naming).
  - Buttons use BS3/4 classes: theme.md demonstrates .btn, .btn-primary, .btn-outline-*, .btn-link, .btn-warning, etc.
[...]

PR #599, opened 2026-02-06
Hugo dart scss

cwhite911:
Switch from css to hugo-dart-scss for styling.

~~Depends on PR #552~~

PR #546, opened 2025-05-29
download: Updated to new download page layout

cwhite911:
I've added the new download layout. However, I need help confirming we have all the content for each OS. @neteler @veroandreo @wenzeslaus @petrasovaa please take a look and let me know if anything is missing. 

We can add missing items to `data/grass.json`

The point of this new layout is to help users install the most current version as quickly as possible for their OS by providing less options.
 
## Linux
image

## macOS
image

## Windows
image

## Docker
image

PR #606, opened 2026-02-28
hot-fix: hamburger dropdown, download buttons, download icons, all news

### LINKED ISSUE #511: Arrows in news page do not fold/unfold content
The arrows next to year numbers do not work, i.e., they do not fold/unfold the news. So, for now, only 2025 news are visible, but they cannot be folded/hidden. 

Image

### LINKED ISSUE #573: Bug Report: Hamburger Menu Not Working on Mobile (Production + Development)
# ** Bug Report: Hamburger Menu Not Working on Mobile (Production + Development)**

## **Description**

The hamburger menu is **not responding in mobile view** on both the:

* **Production site:** [https://grass.osgeo.org/](https://grass.osgeo.org/)
* **Local development environment**
---

## **Observed Behavior**

* Hamburger menu button visually works, but the navigation **does not toggle**.
---

## **Possible Cause**

The project appears to be **primarily built using Bootstrap 4.1.3**, which is located in:

```
static/plugins/bootstrap/
```

However, I found that some templates are using **Bootstrap 5** attributes, such as in file:

```
themes/grass/layouts/partials/footer.html
```

Using mixed Bootstrap versions (4 + 5) **might** be breaking the hamburger toggle functionality.

---

### **Solution 1 (Bootstrap 4 approach)**

Add the required Bootstrap 4 dependencies to the bottom of the navigation template:

```html
<script src="https://code.jquery.com/jquery-3.5.1.slim.min.js"></script>
<script src="https://cdn.jsdelivr.net/npm/popper.js@1.16.1/dist/umd/popper.min.js"></script>
<script src="https://cdn.jsdelivr.net/npm/bootstrap@4.6.0/dist/js/bootstrap.min.js"></script>
```

Bootstrap 4 navbar collapse requires **jQuery + Popper.js** to function correctly.

---

### **Solution 2 (Bootstrap 5 approach)**

Update the navbar HTML to Bootstrap 5 syntax:

```html
data-bs-toggle="collapse"
data-bs-target="#navigation"
```

And ensure Bootstrap 5 JS is loaded correctly.

---

## **Recommendation**

Since the entire project is structured around **Bootstrap 4**, keeping everything consistent with Bootstrap 4 seems safer unless a full migration to Bootstrap 5 is intended.

---
Let me know what you think about this issue. I am also trying to identify the root cause. Please check the problem on the live website. I will start working on a fix as soon as I hear from you and proceed based on your input. Thank you.

### LINKED ISSUE #575: Bug: “Latest New 2025” Dropdown Arrow Points Up When Closed
The dropdown arrow in the **“Latest New 2025”** menu is pointing **up** even when the dropdown is closed.

### **Expected**

* Arrow should point **down** when closed.
* Arrow should point **up** when open.

### **Actual**

* Arrow always points **up**, regardless of state.

Image

I will make PR for it and if any additional changes needed in the Latest News. Let me know and i will make those changes based on your suggestions. Thank you.

### LINKED ISSUE #581: Brand page: download buttons don't work

### LINKED ISSUE #602: Images in the download site are missing
The issue is likely coming from #550, but new approach is needed anyway. The logo can be used as a placeholder. However, the new images can be without the GRASS logo - that is implied, the logos here are for navigation. Something like:

Image

cwhite911:
## Fixes

* Dropdown hamburger button not opening (#573) 
* Dropdown hamburger color (it's now white)
* Brand download buttons (#581) 
* Incorrect path to download page logos (#602)  
* the white gap between navbar and banner on mobile.
* News feed collapse functionally (#511 , #575 ) 

The root issue was caused when we introduced Bootstrap 5 without removing past bootstrap versions and accounting for breaking changes in the bootstrap API.

## Main Issue

In this PR I removed Bootstrap 5 and fixed everything to work with Bootstrap 4.1.3 (the current version in our static plugins directory)

## Testing

I've also included `docker-compose.yml` so you can test the changes locally how they will be rendered on the server. 

To use install docker and then run. 

```
docker compose -f docker-compose.yml up
```

The site will be available at [http://localhost:1313/](http://localhost:1313/).

## Moving Forward

This should resolve most of the major issues with the website at the moment and will give us more time to address PR #599 and issue #600 .

I have 3 main goals I'm suggesting we accomplish.

1. Modernize the websites dependency management to improve consistency between development environments and production. (#552 , #599)
2. Update our dependencies to the latest versions
    * We already introduced renovate which will help  (#591)
3. Switch from css to dart-scss using hugo-extended  (#599)

The biggest challenges are **migrating to Bootstrap 5** (#600) and then we should **migrate from Hugo v0.113.0 to  v0.157.0**. Both of these migrations need planning (Bootstrap issue #600)  as they will break parts of the site, but in my opinion it is worth the effort to use the latest features.

### DIFF
--- a/themes/grass/layouts/download/list.html
+++ b/themes/grass/layouts/download/list.html
@@ -43,7 +43,15 @@
           {{ if not .Params.categories }}
           <div class="section bg-white mb-4">
             <div class="row nm">
-              <div class="col-4 text-center"><img src="{{.Site.BaseURL}}/images/logos/grass-{{.Title}}.jpg" width="110"></div>
+              <div class="col-4 text-center">
+                {{ $imgPath := printf "static/images/logos/historical_logos/grass-%s.jpg" .Title }}
+                {{ if os.FileExists $imgPath }}
+                <img src="{{.Site.BaseURL}}images/logos/historical_logos/grass-{{.Title}}.jpg" width="110" class="pull-right">
+                {{ else }}
+                <img src="{{.Site.BaseURL}}images/logos/grass-logo/grass-green.svg" width="110"
+                  class="pull-right">
+                {{ end }}
+              </div>
               <div class="col-8">
                 <h3 class="mt-2 mb-2">GRASS for {{ .Title }}</h3>
                 <p>{{ .Description }}</p>
@@ -66,7 +74,7 @@
 
           <div class="section bg-white mb-4">
             <div class="row nm">
-              <div class="col-4 text-center"><img alt="GRASS GIS" src="{{.Site.BaseURL}}/images/logos/grass-SampleData.jpg"></div>
+              <div class="col-4 text-center"><img alt="GRASS GIS" src="{{.Site.BaseURL}}images/logos/historical_logos/grass-SampleData.jpg"></div>
               <div class="col-8">
                 <h3 class="mt-2 mb-2">GRASS sample data</h3>
                 <p>Download ready-to-use GRASS sample datasets</p>
--- a/themes/grass/layouts/download/list.html
+++ b/themes/grass/layouts/download/list.html
@@ -44,13 +44,7 @@
           <div class="section bg-white mb-4">
             <div class="row nm">
               <div class="col-4 text-center">
-                {{ $imgPath := printf "static/images/logos/historical_logos/grass-%s.jpg" .Title }}
-                {{ if os.FileExists $imgPath }}
-                <img src="{{.Site.BaseURL}}images/logos/historical_logos/grass-{{.Title}}.jpg" width="110" class="pull-right">
-                {{ else }}
-                <img src="{{.Site.BaseURL}}images/logos/grass-logo/grass-green.svg" width="110"
-                  class="pull-right">
-                {{ end }}
+                <img src="{{.Site.BaseURL}}images/logos/grass-logo-simple/grass-logo-green-simple@1x.png" width="110" class="pull-right">
               </div>
               <div class="col-8">
                 <h3 class="mt-2 mb-2">GRASS for {{ .Title }}</h3>
@@ -63,7 +57,7 @@
           {{ end }}
           <div class="section bg-white mb-4">
             <div class="row nm">
-              <div class="col-4 text-center"><object type="image/svg+xml" data="{{.Site.BaseURL}}/images/logos/grass-logo/grass-green.svg" class="grasslogo"></object></div>
+              <div class="col-4 text-center"><object type="image/svg+xml" data="{{.Site.BaseURL}}/images/logos/grass-logo-simple/grass-logo-green-simple@1x.png" class="grasslogo"></object></div>
               <div class="col-8">
                 <h3 class="mt-2 mb-2">GRASS Addons</h3>
                 <p>Download user contributed GRASS extensions</p>
@@ -74,7 +68,7 @@
 
           <div class="section bg-white mb-4">
             <div class="row nm">
-              <div class="col-4 text-center"><img alt="GRASS GIS" src="{{.Site.BaseURL}}images/logos/historical_logos/grass-SampleData.jpg"></div>
+              <div class="col-4 text-center"><img alt="GRASS GIS" src="{{.Site.BaseURL}}images/logos/grass-logo-simple/grass-logo-green-simple@1x.png"></div>
               <div class="col-8">
                 <h3 class="mt-2 mb-2">GRASS sample data</h3>
                 <p>Download ready-to-use GRASS sample datasets</p>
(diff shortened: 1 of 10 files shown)


cwhite911:
This PR also addresses PR #580

wenzeslaus (changes requested):
The hamburger menu works for me as well as the docker setup I tested. (I did not test with my local non-docker setup, but that does not matter much for this.)

In #603, I actually tried to use the green logo without text from static/images/logos/grass-logo-simple for conda. I think actually made it work only for the page itself. I think that's much better than logos with text and old logos.

grass-logo-green-simple@1x

themes/grass/layouts/partials/navigation.html
```diff
@@ -1,9 +1,9 @@
-<nav class="navbar navbar-expand-lg navbar-light">
+<nav class="navbar navbar-expand-lg navbar-dark">
```
wenzeslaus, line 1: The hamburger is now dark. Is that intentional? Why? ...or is something else going on here?

cwhite911:
> In https://github.com/OSGeo/grass-website/pull/603, I actually tried to use the green logo without text from static/images/logos/grass-logo-simple for conda. I think actually made it work only for the page itself. I think that's much better than logos with text and old logos.

I agree when I was initially doing this I was just fixing the links and wasn't trying to update the design, but it's an easy switch and it will all change anyways once PR #546 is merged.

themes/grass/layouts/partials/navigation.html
cwhite911: What color setting do you have on your monitor and what browser are you using?

petrasovaa: Looks ok (white) for me in chrome.

petrasovaa (dismissed):

themes/grass/layouts/partials/navigation.html
wenzeslaus: I can't reproduce anymore. 

Firefox:

image

Chromium:

image

The color is not exactly white, but works for me in this PR, also my previous concern was in relation to this line which is apparently (navbar-dark) unrelated.

themes/grass/layouts/download/list.html
```diff
@@ -55,7 +57,7 @@ <h3 class="mt-2 mb-2">GRASS for {{ .Title }}</h3>
           {{ end }}
           <div class="section bg-white mb-4">
             <div class="row nm">
-              <div class="col-4 text-center"><object type="image/svg+xml" data="{{.Site.BaseURL}}/images/logos/grass-logo/grass-green.svg" class="grasslogo"></object></div>
+              <div class="col-4 text-center"><object type="image/svg+xml" data="{{.Site.BaseURL}}/images/logos/grass-logo-simple/grass-logo-green-simple@1x.png" class="grasslogo"></object></div>
```
wenzeslaus: Object with SVG-PNG mix shows okay in Firefox, but not Chromium. I guess the root cause here is that we don't have a SVG in this repo.

```suggestion
              <div class="col-4 text-center"><img src="{{.Site.BaseURL}}images/logos/grass-logo-simple/grass-logo-green-simple@1x.png" width="110" class="pull-right"></div>
```

themes/grass/layouts/download/list.html
```diff
@@ -66,7 +68,7 @@ <h3 class="mt-2 mb-2">GRASS Addons</h3>
 
           <div class="section bg-white mb-4">
             <div class="row nm">
-              <div class="col-4 text-center"><img alt="GRASS GIS" src="{{.Site.BaseURL}}/images/logos/grass-SampleData.jpg"></div>
+              <div class="col-4 text-center"><img alt="GRASS GIS" src="{{.Site.BaseURL}}images/logos/grass-logo-simple/grass-logo-green-simple@1x.png"></div>
```
wenzeslaus: Perhaps wrong anyway, and will fixed later, but a consistent image may get rid of the "image too big" at the end.

```suggestion
              <div class="col-4 text-center"><img src="{{.Site.BaseURL}}images/logos/grass-logo-simple/grass-logo-green-simple@1x.png" width="110" class="pull-right"></div>
```

wenzeslaus (commented):
I would hate to postpone this, but only 2 or 3 small fixes to the image would avoid this (Chromium):

Image

1. source: use the same img as elsewhere
2. addons: use img instead of object
3. sample dataset: use img width

themes/grass/layouts/download/list.html
cwhite911: Hmm..this is strange. I'm normalizing them all now, but again this will all change anyways once the new download layout is merged.

cwhite911:
> I would hate to postpone this, but only 2 or 3 small fixes to the image would avoid this (Chromium):

@wenzeslaus Nice catch. I didn't notice the mixing of object and image tags in the list view when I first looked at it.

wenzeslaus (approved):
I did not build the latest version, but assuming there's no typo...

PR #615, opened 2026-04-15
Use Fira Mono or monospace for code, not sans

wenzeslaus:
Fira Sans Mono is not a font name as far as I can tell. The font name is Fira Mono.

Fallback for code should be monospace, not sans-serif, so using that.

We don't ship it or link Fira Mono at this point, but I'm fixing the CSS now as a hot fix and leaving the font addition for later.

I'm setting explicit font for pre so that if pre appears without code and Fira Mono is available, we get consistent font with code.


wenzeslaus:
@cwhite911 Can you please +1 this hotfix, so we have that before the release?

cwhite911 (approved):

PR #635, opened 2026-07-12
Build production site in CI and publish it as a release artifact

cwhite911:
## What this does

Adds a CI workflow that builds the website on every push to `master` (Hugo Extended + Dart Sass + `npm ci` when present) and publishes `public/` as a checksummed tarball on a rolling `production` release:

```
https://github.com/OSGeo/grass-website/releases/download/production/grass-website.tar.gz
https://github.com/OSGeo/grass-website/releases/download/production/grass-website.tar.gz.sha256
```

## Why

The upcoming Bootstrap 5 / Dart Sass migration (#600, starting with #599) changes the build requirements to Hugo **Extended** + `npm ci` + an external **Dart Sass** binary. The production cronjob on grass.osgeo.org currently builds the site on the server with a plain `go install`-ed Hugo and would fail after the migration merges.

With this workflow, the server never builds the site: its cronjob just downloads the tarball, verifies the SHA256, and rsyncs it into the web root, so the toolchain stays pinned in CI only. No token is needed on the server (public release URL over HTTPS), which also avoids the expiring-PAT caveat of the existing GHA-artifact fetch used for the manuals. Companion server-side change: `grass-addons` `utils/cronjobs_osgeo_lxd/hugo_clean_and_update_job.sh` (PR to follow).

## Design notes

- The npm and `.nvmrc` steps are guarded, so the workflow builds both the current master (no `package.json`) and the post-migration branches.
- The pinned Hugo version is read from `.env` when present (the migration adds it); until then it defaults to `0.113.0`. Note `v0.113.8` (pinned in the preview `hugo.yml` on the #599 branch) does not exist as a Hugo release; `0.113.0` matches the production server.
- Uses the built-in `GITHUB_TOKEN` with `contents: write` only; the release is marked prerelease so it never appears as "Latest release".
- Every merge to master overwrites the release assets (`--clobber`); `BUILD_INFO` inside the tarball records the commit, timestamp, and run URL.

## Tested

Ran end-to-end on my fork via `workflow_dispatch` against current master content:
- all steps green; release + both assets published
- `sha256sum -c` passes on a downloaded copy
- tarball unpacks to 234 HTML pages; sitemap page count matches the live site exactly (228 == 228)

Part of the deployment plan for #600.

### DIFF
--- a/.github/workflows/build-production-site.yml
+++ b/.github/workflows/build-production-site.yml
@@ -33,27 +33,34 @@ jobs:
   build:
     runs-on: ubuntu-latest
     env:
-      # Fallback only: .env is the source of truth for the pinned Hugo version
-      # once present (see README "For website developers").
+      # Fallbacks only: .env is the source of truth for the pinned toolchain
+      # versions once present (see README "For website developers").
       HUGO_VERSION_DEFAULT: 0.113.0
+      DART_SASS_VERSION_DEFAULT: 1.101.0
     steps:
       - name: Checkout
         uses: actions/checkout@v6
 
-      - name: Determine Hugo version (.env is the source of truth)
-        id: hugo
+      - name: Determine toolchain versions (.env is the source of truth)
+        id: versions
         run: |
-          # Extract only the value; do not source .env (that would execute
+          # Extract only the values; do not source .env (that would execute
           # arbitrary shell content).
+          get_env_value() {
+            grep -E "^$1=" .env 2>/dev/null | head -n1 | cut -d= -f2- | tr -d '"' | tr -d "'"
+          }
           HUGO_VERSION=""
+          DART_SASS_VERSION=""
           if [[ -f .env ]]; then
-            HUGO_VERSION="$(grep -E '^HUGO_VERSION=' .env | head -n1 | cut -d= -f2- | tr -d '"' | tr -d "'")"
+            HUGO_VERSION="$(get_env_value HUGO_VERSION)"
+            DART_SASS_VERSION="$(get_env_value DART_SASS_VERSION)"
           fi
-          echo "version=${HUGO_VERSION:-$HUGO_VERSION_DEFAULT}" >> "$GITHUB_OUTPUT"
+          echo "hugo=${HUGO_VERSION:-$HUGO_VERSION_DEFAULT}" >> "$GITHUB_OUTPUT"
+          echo "dart_sass=${DART_SASS_VERSION:-$DART_SASS_VERSION_DEFAULT}" >> "$GITHUB_OUTPUT"
 
       - name: Install Hugo CLI (extended)
         env:
-          HUGO_VERSION: ${{ steps.hugo.outputs.version }}
+          HUGO_VERSION: ${{ steps.versions.outputs.hugo }}
         run: |
           # Hugo release assets have used two naming schemes; try both.
           wget -O ${{ runner.temp }}/hugo.deb \
@@ -68,7 +75,7 @@ jobs:
       # planned Hugo/Dart Sass upgrade inherits a reproducible toolchain.
       - name: Install Dart Sass
         env:
-          DART_SASS_VERSION: 1.101.0
+          DART_SASS_VERSION: ${{ steps.versions.outputs.dart_sass }}
         run: |
           curl -fsSL -o "${{ runner.temp }}/dart-sass.tar.gz" \
             "https://github.com/sass/dart-sass/releases/download/${DART_SASS_VERSION}/dart-sass-${DART_SASS_VERSION}-linux-x64.tar.gz"
--- a/.github/workflows/build-production-site.yml
+++ b/.github/workflows/build-production-site.yml
@@ -61,13 +61,14 @@ jobs:
       - name: Install Hugo CLI (extended)
         env:
           HUGO_VERSION: ${{ steps.versions.outputs.hugo }}
+          TEMP_DIR: ${{ runner.temp }}
         run: |
           # Hugo release assets have used two naming schemes; try both.
-          wget -O ${{ runner.temp }}/hugo.deb \
+          wget -O $TEMP_DIR/hugo.deb \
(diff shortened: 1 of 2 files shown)


echoix:
Remember if it is the server side that needs to pull  the tarball, it should fail gracefully if the artifact is expired and can’t be downloaded. In some way, it would be better to push the tarball when a new one is available rather than pulling it, but there’s different things to take care about security-wise in both cases

cwhite911:
@echoix Expired artifacts is a property of Actions artifacts (90-day retention), and this workflow publishes to a release asset instead, which never expires. The rolling production release always holds the latest build at a stable URL. That's also why I went with a release rather than the artifact-download pattern already used for the manuals (gh_cli_download_artifact.sh), which additionally needs a PAT that expires every 3 months.

The companion cronjob (OSGeo/grass-addons#1783) fails gracefully. In the worst case the site stays on the last good build. I agree there are trade-offs using push verses pull, push would require CI to hold a credential that can write to grass.osgeo.org (deploy key or a webhook receiver), so a compromised workflow would have direct write access to the production web root. Pull needs zero credentials in either direction which keeps the trust root identical to the current pull-based cronjob and matches the existing pull-based fetch the server already does for the manuals. The only cost is deploy latency, bounded by the same 12-hour cron cadence production uses today.

.github/workflows/build-production-site.yml
```diff
@@ -0,0 +1,135 @@
...
+    shell: bash
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    env:
+      # Fallbacks only: .env is the source of truth for the pinned toolchain
+      # versions once present (see README "For website developers").
+      HUGO_VERSION_DEFAULT: 0.113.0
+      DART_SASS_VERSION_DEFAULT: 1.101.0
```
echoix: We will probably want to have renovate keep those updated; They will diverge at some point from what is actually used.

On second thought, we would prefer have it fail and only use one source of truth. The .env versions still need to stay updated

.github/workflows/build-production-site.yml
```diff
@@ -0,0 +1,135 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
```
echoix: It is now v7

.github/workflows/build-production-site.yml
```diff
@@ -0,0 +1,135 @@
...
+          fi
+          echo "hugo=${HUGO_VERSION:-$HUGO_VERSION_DEFAULT}" >> "$GITHUB_OUTPUT"
+          echo "dart_sass=${DART_SASS_VERSION:-$DART_SASS_VERSION_DEFAULT}" >> "$GITHUB_OUTPUT"
+
+      - name: Install Hugo CLI (extended)
+        env:
+          HUGO_VERSION: ${{ steps.versions.outputs.hugo }}
+        run: |
+          # Hugo release assets have used two naming schemes; try both.
+          wget -O ${{ runner.temp }}/hugo.deb \
```
echoix: Template injection

.github/workflows/build-production-site.yml
```diff
@@ -0,0 +1,135 @@
...
+            "https://github.com/gohugoio/hugo/releases/download/v${HUGO_VERSION}/hugo_extended_${HUGO_VERSION}_Linux-64bit.deb" \
+          || wget -O ${{ runner.temp }}/hugo.deb \
```
echoix: Template injection

.github/workflows/build-production-site.yml
```diff
@@ -0,0 +1,135 @@
...
+            "https://github.com/gohugoio/hugo/releases/download/v${HUGO_VERSION}/hugo_extended_${HUGO_VERSION}_linux-amd64.deb"
+          sudo dpkg -i ${{ runner.temp }}/hugo.deb
```
echoix: Template injection

.github/workflows/build-production-site.yml
```diff
@@ -0,0 +1,135 @@
...
+
+      # Pinned for reproducibility. Note: with Hugo 0.113 the site compiles via
+      # Hugo Extended's built-in libsass (head.html sets no "transpiler"
+      # option), so this binary is not exercised yet; it is pinned now so the
+      # planned Hugo/Dart Sass upgrade inherits a reproducible toolchain.
+      - name: Install Dart Sass
+        env:
+          DART_SASS_VERSION: ${{ steps.versions.outputs.dart_sass }}
+        run: |
+          curl -fsSL -o "${{ runner.temp }}/dart-sass.tar.gz" \
```
echoix: Template injection

.github/workflows/build-production-site.yml
```diff
@@ -0,0 +1,135 @@
...
+            "https://github.com/sass/dart-sass/releases/download/${DART_SASS_VERSION}/dart-sass-${DART_SASS_VERSION}-linux-x64.tar.gz"
+          tar -xzf "${{ runner.temp }}/dart-sass.tar.gz" -C "${{ runner.temp }}"
```
echoix: Template injection

.github/workflows/build-production-site.yml
```diff
@@ -0,0 +1,135 @@
...
+          echo "${{ runner.temp }}/dart-sass" >> "$GITHUB_PATH"
```
echoix: Template injection

echoix (changes requested):

echoix:
The .env file is not commited. Usually, .env is always in .gitignore, and only a template, or .env.example or something like that is committed. But they are probably other ways to define the tooling used and versions used for web projects. What is hugo working well with?

cwhite911:
> The .env file is not commited. Usually, .env is always in .gitignore, and only a template, or .env.example or something like that is committed. But they are probably other ways to define the tooling used and versions used for web projects. What is hugo working well with?

Let's merge this without any .env file. So that I can merge it into PR #599 which has the rest of the context. I will remove the commited .env in favor of letting renovate handle the version updates.

echoix (approved):

Issue #636, opened 2026-07-13
Add zizmor to the checks run


echoix:
Add zizmor either on CI or through pre-commit, if it is used here, to catch if some bad patterns are being used in workflow files

PR #645, opened 2026-07-14
Single-source brand tokens; remove orphaned style.css (issue #600)

cwhite911:
> **Draft: stacked on #599 → #644.** The diff includes both parents' commits until they merge; the two commits to review here are "Fix code-block highlighting…" and "Single-source brand tokens…". Will be marked ready once #644 lands.

## What this does

Makes the SCSS the single source of truth for the GRASS brand tokens, and fixes two regressions surfaced by the migration (part of #600):

- **`_css-variables.scss`**: generates the `:root { --gs-* }` custom-property block from the `$gs-*` Sass tokens (each light/lighter/lightest ramp derived from its base color), so the served stylesheet finally defines the `var(--gs-*)` names the brand page documents. 73 custom properties emitted.
- **Deletes the orphaned `themes/grass/assets/css/style.css`** (2000 lines): hand-maintained, no longer served since #599, and already drifted from the SCSS. Its post-fork hot-fixes were ported into the SCSS in #599.
- **Reconciles token drift** at the source (primary-light/secondary-light ramp typos, support-color alpha) and fixes the brand page's `--gs-primary-ligh-color` typo.
- **Code-block highlighting fix**: re-links the highlight.js stylesheet (dropped in the migration) and disables Hugo's Chroma (`markup.highlight.codeFences = false`) so fenced code is no longer double-highlighted into dark-on-dark.
- **Timeline margin fix**: `mx-0` on the bare `.row` wrappers so Bootstrap 5's negative gutters stop squashing the releases/web-evolution pages.

## Verified

- Clean containerized build; the compiled CSS now contains the full `--gs-*` token set and tokens resolve at runtime on the brand page.
- Headless-browser QA 15/15; the new pixi instructions on the conda page render through highlight.js with no Chroma remnants.

Part of #600. Follow-up after this lands: export the same token block to the GRASS core docs (mkdocs/pygrass) and Quarto tutorials to replace today's hand-copied duplicates.

PR #644, opened 2026-07-14
Migrate markup to Bootstrap 5 and drop jQuery (issue #600)

cwhite911:
> **Stacked on #599** — the diff below includes #599's commits until it merges. The commits to review here start at "Migrate markup to Bootstrap 5".

## What this does

Completes the Bootstrap 3 to 5 markup and JavaScript migration (part of #600), on top of the Dart Sass foundation in #599:

- **`data-*` attribute sweep**: `data-toggle` / `data-target` / `data-dismiss` / `data-ride` / `data-interval` to `data-bs-*` across nav, tabs, dropdowns, carousel, modal, and collapses (125 attributes across layouts and content).
- **History timelines** (`about/history`, `releases`, `web-evolution`): converted to independent Bootstrap 5 collapses (`data-parent` is removed in BS5), `.in` to `.show`, `.sr-only` to `.visually-hidden`, and a rotating chevron affordance on each panel title. The custom timeline design is preserved.
- **Utility renames**: `mr-*`/`ml-*` to  `me-*`/`ms-*`, `pull-right` to `float-end`, `img-responsive`  to `img-fluid`; modal close to `.btn-close`; fixed a duplicate tab `id`.
- **Dropped jQuery entirely**: `script.js` rewritten in vanilla JS (hero offset, link tagging, gallery toggle, OS-detect via the BS5 Tab API), the copy-to-clipboard partial ported, `bootstrap.bundle.min.js` loaded standalone with SRI, and the `jquery` dependency removed from `package.json` and `config.toml`.
- Visual fixes found in review: centered home topic-card icons, history collapse affordance.

## Verified

- Clean containerized build (Hugo 0.113.0 Extended + npm ci), zero legacy `data-toggle`/`data-parent` in the rendered site, no jQuery `<script>` anywhere.
- Headless-browser QA (15/15): no JS errors on key pages; nav dropdown, carousel, donate modal, brand download dropdowns, and history collapses all work, including collapse independence.

Part of #600. Next in the stack: brand-token single-sourcing (draft PR to follow).

PR #646, opened 2026-07-14
Split production artifact workflow into build and deploy jobs

cwhite911:
## What this does

Splits the production artifact workflow into separate `build` and `deploy` jobs:

- **`build`** (read-only token): checkout, toolchain install, `hugo --minify`, `BUILD_INFO`, tar + sha256, then hands the tarball to the next job as a short-lived (1 day) workflow artifact.
- **`deploy`** (`needs: build`, the only job with `contents: write`): downloads the artifact, **re-verifies the checksum**, moves the rolling `production` tag to the built commit, and uploads the release assets.
- Adds Zizmor job to address ( #636 )

## Why

- A failed build can never reach the publish step: `needs: build` skips `deploy` entirely, so the release assets always stay on the last good build.
- Least privilege: the bulk of the workflow (all the toolchain and build steps) now runs with a read-only token; only the small publish job can write.
- The checksum re-verification also guards against any corruption in the artifact handoff itself.
- Clearer Actions UI: build failures and publish failures are visibly distinct.

## Tested

End-to-end on a fork via `workflow_dispatch`: both jobs green, artifact handoff and checksum verification pass, the `production` tag advanced to the built commit, and the downloaded release assets verify with `sha256sum -c`. zizmor reports no findings.

No change to the published artifact format or URLs; the server-side cronjob (OSGeo/grass-addons#1783) is unaffected.


cwhite911:
Added a zizmor workflow-audit job (zizmor-action, SARIF upload to code scanning) per review discussion. Heads up: on first run it will flag four findings in the current `hugo.yml` (workflow-level `pages`/`id-token` write permissions, credential persistence, one template expansion). Those are all already fixed on the #599 branch and will clear when it merges.

.github/workflows/build-production-site.yml
```diff
@@ -101,14 +113,43 @@ jobs:
...
+          path: |
+            grass-website.tar.gz
+            grass-website.tar.gz.sha256
+          retention-days: 1
+          if-no-files-found: error
+
+  deploy:
+    # Pull requests only validate the build; publishing happens on pushes to
+    # master (and manual dispatch).
+    if: ${{ github.event_name != 'pull_request' }}
```
echoix: Also limit to the main branch, probably double check if using environments and environment protection rules (not sure about the naming) would allow to add restrictions on what branches (or conditions need to be made) to "use" or "have access" (or "acquire the lock named '(environment name)'") can really deploy.

Using environments to gate deployments can also be used to add things like a number of approvals before deploying/publishing.

.github/workflows/build-production-site.yml
```diff
@@ -6,23 +6,33 @@
...
-  contents: write
+# Each job declares exactly what it needs.
+permissions: {}
 
-# A newer master build supersedes an in-flight one.
+# A newer build of the same ref supersedes an in-flight one; the ref-scoped
+# group keeps pull request builds from cancelling master deployments.
 concurrency:
-  group: production-site
+  group: production-site-${{ github.ref }}
```
echoix: Doublecheck that it doesn't allow double up the jobs that could run, or cancel unwanted jobs once merged. If you want the same behaviour than on main branch of the main repo, we really need to use the same syntax.

If you want to only allow one per PR, but build every merged PR separately, queuing every one without any cancellation, this change doesn't do that.

Maybe you want, on the default branch, to cancel not started jobs; so if one is already in progress + 3 queued ones + a 5th one,  you would keep the in progress, cancel three ones in the middle, and queue the 5th one.
But with this, that is probably what `group: production-site-${{ github.ref }}`, you would end up that if while the previous scenario is queued, and you do a rerun of a previous commit, it will end up cancelling the "5th" that was the only non-cancelled job that would run. We end up having the "previous" commit rerun to be the "latest" to win and end up deployed to the server at the end.

A third element in this discussion: You would probably want to use environments in this workflow for another reason: to serialize the execution of jobs, even if you would have access to 20 commits built in parrallel. Especially for the deploy step, you wouldn't want for the slowest job run (the one that took 3 seconds more) to win and be the last one to push and that would get deployed. They need to be "serialized", and the tool for it is most probably environments.

.github/workflows/zizmor.yml
echoix, line 1: I discussed with @cwhite911, and while this zizmor workflow is correct, fine, and does the right thing, we shouldn't keep it long term. We would want to make it run locally too. If there's no common mechanism for everyone to run the same thing, we end up with a "is fine on my machine" situation.

We could do it like the main repo, or grass-addons, with pre-commit, but if we don't enforce it in CI, it won't be kept "green". Since it is useful to also have suggestions posted as review comments when possible, we then end up needing to do the same thing as grass-addons repo.

So the plan is:

1. Add a .pre-commit-config.yaml file with at least zizmor. Get insppired by the main repo or the grass-addons (they should be very similar, they got synced as much as possible not too long ago). Adjust excluded files.
2. From grass-addons repo, copy the `.github/workflows/additional_checks.yml`.
3. Adjust branches and triggers to `additional_checks.yml` (compare with the one on the main grass repo to see what changes).
4. From the grass-addons repo, copy the `.github/workflows/post-pr-reviews.yml` file. Adjust the name of the workflows to monitor.
5. From the grass-addons repo, copy the `zizmor.yml` config file. It ignores the fact that we are using an unpinned version of an action, because we maintain it from the main repo.
6. (Maybe copy the `biome.jsonc` config file from the grass-addons repo if wanted)
7. From the grass-addons repo, copy the section of the `renovate.json5` config file that stops wanting to pin the reusable workflow and rebases the PR on every new commit of the main repo. If you'd want the website to be fully reproducible to a previous state of the commit, you could allow to pin and decide on a schedule on when to update the SHA of the main repo.
8. Make the additionnal_checks workflow required in the repo's settings.
9. Consider a new name for additionnal checks. Its now kind of weird, and finally a required one in the main repo, now that we replaced previous tools that duplicated checks into only a pre-commit version.

.github/workflows/build-production-site.yml
echoix: An idea for this here, what about keeping at least a couple of previous versions, and maybe major historic ones. It could be on the server though, but I thought about it in a tags/release way.
- In case of a bad file/bad build, we could manually define a previous release as the latest, and get the server to download it, by using the "latest" release url shortcut: https://docs.github.com/en/repositories/releasing-projects-on-github/linking-to-releases#linking-to-the-latest-release
- If a server script is bad, or lose the server, we could upload to a new infra the known working version easily.
- If it was done with new "calendar-dated" tags, thus updating often, people can unsubscribe to releases for this repo, so it doesn't matter to not overwrite.
- Keep assets named the same, so the server always looks for the same artifact to download OR have the names change so that the server natively has backups downloaded (I dont like this idea though, too complex).
- At one point I dont know if immutable releases will be forced upon everywhere, but it is not a so bad idea too. But if it does, not overwriting would be useful.

Do we have any limits on the release artifacts total size across releases? A script to go and clean in the backs, like a backup retention tool does?

echoix (changes requested):
Couple of suggestions. The deploy part and concurrency group usage only is not enough to handle the cases we want, especially if there's more than one deploy job (let's say 20 jobs) try to deploy at once. Which one should be the one that wins? The current workflow doesn't manage this properly or at all. (The answer would be: the slowest one, so the last to finish, not the latest one we would expect to have).

.github/workflows/zizmor.yml
cwhite911: Implemented your plan: 

1. pre-commit config aligned with grass-addons (adjusted excludes for Hugo templates and the content the migration stack rewrites)
2. Additional Checks running prek with the suggestions upload
3. post-pr-reviews posting the pre-commit diff
4. your zizmor.yml policy for create-upload-suggestions, and the Renovate rule to keep it unpinned at main. 

The standalone zizmor workflow is gone, replaced by the pre-commit hook;
zizmor.yml carries temporary ignores for the pre-migration hugo.yml that disappear once #599 lands. I skipped biome.jsonc for now since it would reformat the JS that #644 rewrites; that plus widening the markdownlint scope to content/ is the post-stack cleanup.

Two open items from your list for an admin: making Additional Checks required on master, and whether to rename it.

cwhite911:
@echoix once this PR is approved and merged I'm ready to merge the remaining compontents to deploy the website.

GRASS website
- #599 
- #644 
- #645 
- [https://github.com/OSGeo/grass-addons/pull/1783](https://github.com/OSGeo/grass-addons/pull/1783)
- #600

petrasovaa (approved):

.github/workflows/build-production-site.yml
```diff
@@ -101,28 +121,110 @@ jobs:
...
+            | jq -r --argjson keep "$KEEP" '
+                [.[] | select(.tagName | startswith("site-"))]
+                | sort_by(.createdAt) | reverse | .[$keep:]
+                | .[] | select(.name | test("\\[keep\\]") | not) | .tagName'
+          )
+          for tag in "${stale[@]}"; do
+            echo "Pruning old release $tag"
+            gh release delete "$tag" --repo "$GITHUB_REPOSITORY" --cleanup-tag --yes
+          done
+          echo "Pruned ${#stale[@]} release(s); kept newest $KEEP plus any [keep]."
```
echoix, line 230: In the future, you might want to have the prune be a separate job, on schedule. A new iteration is not needed,

.github/workflows/build-production-site.yml
```diff
@@ -101,28 +121,110 @@ jobs:
...
+          head_sha="$(gh api "repos/$GITHUB_REPOSITORY/git/ref/heads/$GITHUB_REF_NAME" --jq '.object.sha')"
+          if [[ "$head_sha" == "$GITHUB_SHA" ]]; then
+            echo "current=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "current=false" >> "$GITHUB_OUTPUT"
+            echo "::notice::Skipping publish: $GITHUB_REF_NAME moved to ${head_sha};" \
+              "a newer run owns the publish (this run built ${GITHUB_SHA})."
           fi
-          gh release upload production \
+
```
echoix, line 170: ~I'm not 100% sure. That's what I understand could be happening:~

~1. Many (e.g. 30) new commits are merged, and the last couple of them (e.g. 3) are spaced out by let's say 30 secs.
2. The queue is full, all "build" jobs are loading the "deploy" queue, serialized by the environment.
3. Lets start looking at the 27th. Since it was a bit delayed, it doesn't start immediatly, but the "build" artifact for the 28th doesn't exist yet. 
4. The 27th artifact is the "latest" artifact available, both in time, and topologically (no new commits).
5. Ignoring the concurrency race issue that might happen because the check is as the beginning of the run (anyways, it "locks" using an environment), the 28th commit might arrive just before doing the check.
6. The check runs, and would be false, even if the 28th commit would run later. Maybe that's what you intended here.
7. At the end of that spike, the 30th should win.
8. If you finally decide to workflow dispatch (not specifying a branch)....~

oh wait, I rubber-ducked myself when rereading the current environment and deployment docs, and understood more....

Since the concurrency group is on the whole group of two jobs, it should not lead to a situation where the two jobs are separated.
Oh, but the environment is only on the second job...


Anyways, I tried locally the `gh api` command, and if you create a tag to any commit you want, you can easily get the workflow dispatch to run on that tag and get the condition to be true, so to not use when there are other jobs in master that are not deployed yet in order to not mess up the "order" .

echoix (approved):
I wasn't sure at first, as I thought I found a loophole, but when explaining it I found that it might be correct. Let's just observe if it behaves correctly, by keeping an eye at that section?

PR #689, opened 2026-08-03
Templates: replace deprecated Hugo APIs, drop Babel, prune config

cwhite911:
Removes every template construct that errors on current Hugo, so #638 gets closer to a version-pin flip. Three commits:

1. **Babel removal.** The photoswipe lightbox was transpiled for old browsers and then loaded with `type="module"`, which those browsers ignore. The module is now served minified and fingerprinted; `@babel/*` devDependencies dropped.
2. **Config hygiene.** `security.exec.allow` reduced to `['^go$', '^sass$']` (nothing invokes dart-sass-embedded, postcss, babel, or npx; `^sass$` pre-arms the Dart Sass switch). Removes the misplaced `enableInlineShortcodes` under `[security]` and two mounts with nonexistent sources.
3. **Template APIs.** `.Data.Pages`, unguarded `.File.UniqueID`, `.Site.Author`, `.Site.Config.Services.RSS`, and the global `minify`/`fingerprint` aliases replaced with forms valid on both 0.113 and current Hugo (`toCSS` stays: `resources.ToCSS` is removed upstream, `css.Sass` does not exist on 0.113). `Site.BaseURL` string concatenation emitted `//` URLs on every page; now `relURL`. News listings use `site.RegularPages`, which also stops the news section page listing itself as a 2019 item.

Verified on Hugo 0.113 (production compose build; output diff limited to the intended URL fixes) and on Hugo 0.164 extended, which now builds with zero errors. Remaining 0.164 items are upgrade-time only: `_build` front matter rename in six download pages, `allowContent` for the four 1996 HTML news files, `languageCode`/`languageName`/`.Site.Data` renames (their replacements do not exist on 0.113), and the sitemap output format.

Part of the cleanup stack preparing #638.


cwhite911:
Superseded by the stacked-PR version of this branch (same commits, same-repo branch so it can be stacked); link follows.

cwhite911:
Resubmitted as #694, bottom of stack #696.

echoix:
> Superseded by the stacked-PR version of this branch (same commits, same-repo branch so it can be stacked); link follows.

Ok, it doesn't work for PRs from forks... :( Maybe that's why I wasn't able for the main repo to get it working in San-Michele. I didn't find why until now.

Issue #670, opened 2026-07-18
Table layout needs revision


veroandreo:
The style for tables needs revision. I built #665 with all new styling now, and this is how the table there looks like:

<img width="600" alt="Image" src="https://github.com/user-attachments/assets/332201fb-41ff-4633-b26b-620157dc3147" />

cwhite911:
> I built https://github.com/OSGeo/grass-website/pull/665 with all new styling now, and this is how the table there looks like:

I just noticed this part. Check the styling with the change from PR #703 and let me know if you prefer it another way. I was trying to fix the generic base table and not a special view.

PR #698, opened 2026-08-07
SCSS: remove dead selectors, files, and token variables

cwhite911:
Removes the SCSS with no consumers, verified by scripted grep across layouts, content, data, config, and JS, then re-checked in the built site:

- the search-form and jQuery-UI partials (both features removed long ago)
- 56 dead rule blocks (unused brand color utilities, an unused spacing ladder, leftover component styles)
- the shadowed `body`/`.grass-bg-image` duplicates in `main.scss`
- 39 unreferenced token variables

Kept deliberately, with a new README note explaining why grep alone cannot judge them: runtime-injected families (highlight.js, OpenLayers, script.js), Bootstrap-generated classes and variable overrides, the published `--gs-*` token API, and classes referenced only from raw HTML in content (for example the history pages' `.glyphicon` markup, which an earlier audit had wrongly called dead).

Verified on Hugo 0.113 (compose production build): `css/style.css` is the only output file that changes (5.8 kB smaller); every removed class is absent from the built HTML and JS. Compiles clean under Dart Sass.

Part of stack #700, preparing #638.


cwhite911:
@veroandreo will you review/approve 

- #699
- #698 

It will unblock

- #670

veroandreo (approved):

cwhite911:
Thank you @veroandreo

wenzeslaus (approved):
I didn't check the details, but looks reasonable to me.

PR #703, opened 2026-08-19
scss: add markdown table render hook, and fallback for pre hugo >=0.134

### LINKED ISSUE #670: Table layout needs revision

cwhite911:
Addresses issue #670

Screenshot from 2026-08-19 01-01-00


echoix (approved):
Maybe Vero would like to see it before merging

Issue #678, opened 2026-07-23
Revise Learn GRASS Page


baharmon:
I suggest revising the index page for Learn GRASS. The Learn GRASS dropdown menu in the header works great, but the index page is inconsistent and thus confusing to navigate.

The Learn GRASS dropdown menu in the header of https://grass.osgeo.org/ has the following entries:

* General overview
* First time users
* Documentation
* Tutorials
* Gallery
* Books
* Try online

The index page for Learn GRASS, however, has links for:

* Manual pages (ie. docs)
* Wiki
* Courses & workshops (but link is for tutorials)
* Notebooks (should add [Python Intro](https://grass.osgeo.org/grass-stable/manuals/python_intro.html))
* Videos & presentations (but no links!)
* Books
* Google Scholar
* GRASS Online

Perhaps the [index page](https://grass.osgeo.org/learn/) and [overview](https://grass.osgeo.org/learn/overview/) should be merged. The overview is introductory and more useful. The index page doesn't have much content except links which are inconsistent with the dropdown list in the header and have outdated language. I suggest the descriptive paragraph be follow by a bullet list of links. The icons and videos on the index page feel cluttered and improperly prioritized. For example, the video embed should probably be for the [GRASS Youtube channel](https://www.youtube.com/@grass-gis) rather than just the NCSU GeoForAll Lab channel. The tutorials at https://grass-tutorials.osgeo.org/ should be prioritized here, but are mislabeled (as courses), don't have an icon, and thus are hard to find.

Note: The tutorials page https://grass-tutorials.osgeo.org/ is titled Learn GRASS just like https://grass.osgeo.org/learn/ but has different content. This is also inconsistent and confusing for navigation. Given the naming, it feels like the latter should just link to the former, except of course there is a lot of other content like docs, books, and videos to showcase.

PR #711, opened 2026-08-31
learn: Update learning page content and redesign

cwhite911:
I've updated the https://grass.osgeo.org/learn/ page with updated content and a new layout.

## Current Page
image

## New Page
image

It could use an updated YouTube video for the "What is GRASS" and we could curate the content as needed.

Related to: https://github.com/OSGeo/grass-website/issues/678


wenzeslaus (commented):
I think it should use even more project-maintained resources, so instead of the NCSU GeoForAll Lab channel, I linked playlist from the GRASS channel. I made the changes directly as well: List of notebooks is a good idea, but we don't have a good one. The wiki is not great, so I replaced that with Binder opened in a dir for now. The featured content for a video now has the actual video, more to get an image than anything else. There were two Try online links and now there would be more or less three, so I replaced one with beginners link.

Image

petrasovaa (approved):
Looks good!

